PR #63, opened 2016-06-22
Allow extra data to be carried in Request and Permission

joeconwaystk:


anachlas:
👍

lib/auth/protocols.dart
```diff
@@ -54,6 +54,11 @@ class Permission {
 
   /// The [AuthenticationServer] that granted this permission.
   final AuthenticationServer grantingServer;
+
+  /// Container for any data a [RequestHandler] wants to attach to this permission for the purpose of being used by a later [RequestHandler].
+  ///
+  /// Use this property to attach data to a [Request] for use by later [RequestHandler]s.
```
anachlas: Noticed after I merged, but this should be Permission instead of Request right?

joeconwaystk: Yeah, we can fix that later I guess, the docs are far from complete :)

PR #65, opened 2016-06-29
Reuse refresh token for new access token

smswz:


lib/auth/authentication_server.dart
```diff
@@ -83,17 +83,24 @@ class AuthenticationServer<ResourceOwner extends Authenticatable, TokenType exte
...
     token.resourceOwnerIdentifier = ownerID;
     token.clientID = clientID;
 
     return token;
   }
 
+  /// Given a [TokenType], generates and sets a new access token.
+  ///
+  /// This updates the access token, issue and expiration dates for a token.
+  void appendNewAccessToken(TokenType token, int expirationInSeconds) {
```
joeconwaystk: I don't care for the name of this method.. I don't have a better idea though. I am also not all that sure it needs to be abstracted into a new method, given its brevity

smswz: I didn't want to have the expiration calculation in two places, but I also hate the name and couldnt come up with anything better. I'll just duplicate it in both places. I'll save the function call for when there are 3x uses

joeconwaystk: Why not make a method just for calculating an expiration date? DateTime expirationDateForSeconds(DateTime issueDate, int timeInSeconds)

smswz: A one line function seems a bit extreme

joeconwaystk: Perfectly reasonable to have one line calculation functions to ensure consistency across call sites. Having a well named method like that makes it easy to reason about your code, this current method is nebulous - append to what? - and is just a glob of side effects. I can't determine from the call site whether or not is actually does something - like maybe it calls the delegate storage method internally. With code like this:

```
token
  ..refreshToken = randomStringOfLength(32)
  ..accessToken = randomStringOfLength(32)
  ..issueDate = new DateTime.now().toUtc()
  ..expirationDate = expirationDateForToken(token, expirationInSeconds);
```

The code all flows together, and it's clear we're assigning a bunch of properties, esp. with the added clarity of cascading.

joeconwaystk: That, or something on the Token itself for setting an issueDate and a duration at the same time. That may be better...

```
token
  ..refreshToken = randomStringOfLength(32)
  ..accessToken = randomStringOfLength(32)
  ..setTimestamps(new DateTime.now().toUtc, expirationInSeconds);
```

Issue #39, opened 2016-01-26
Replace request body parser


joeconwaystk:
Currently uses outdated http_server, which has an issue of responding to requests when it fails, which  makes no sense.

PR #68, opened 2016-07-02
Jc/replacebody

joeconwaystk:
Removes some warnings.

Removes http_server dependency.

Implements HTTP request body decoding.

Finishes #39

### DIFF
--- a/lib/base/body_decoder.dart
+++ b/lib/base/body_decoder.dart
@@ -1,0 +1,115 @@
+part of aqueduct;
+
+/// A decoding method for decoding a stream of bytes into a String based.
+///
+/// Typically, this represents the function [Utf8Codec.decodeStream].
+typedef Future<String> BodyStreamDecoder(Stream s);
+
+class HTTPBodyDecoder {
+  /// The set of available decoders for HTTP request data.
+  ///
+  /// The HTTP header for Content-Type is broken into two pieces, primary type (e.g., 'application') and subtype (e.g., 'json').
+  /// The primary type is the first key in [decoders], and then the subtype is used to return the specific encoding
+  /// function from this map. Decoders take a [HttpRequest] and return the decoded value of the request body based on the Content-Type pair.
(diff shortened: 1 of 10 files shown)


lib/base/http_controller.dart
```diff
@@ -90,25 +90,11 @@ abstract class HTTPController extends RequestHandler {
...
-      if (matchingContentType == null) {
-        throw new _InternalControllerException("Unsupported Content-Type", HttpStatus.UNSUPPORTED_MEDIA_TYPE);
-      }
-
-      try {
-        await req.decodeBodyWithDecoder((r) async {
-          return (await HttpBodyHandler.processRequest(r)).body;
-        });
-      } catch (e) {
-        throw new _InternalIgnoreBullshitException();
```
anachlas: 👍

lib/base/body_decoder.dart
```diff
@@ -0,0 +1,115 @@
...
+  /// The HTTP header for Content-Type is broken into two pieces, primary type (e.g., 'application') and subtype (e.g., 'json').
+  /// The primary type is the first key in [decoders], and then the subtype is used to return the specific encoding
+  /// function from this map. Decoders take a [HttpRequest] and return the decoded value of the request body based on the Content-Type pair.
+  /// The following decoders are supported by default (you may add extra decoders using [addDecoder], do not manipulate [decoders] directly.)
+  ///
+  ///       application/json
+  ///       application/x-www-form-urlencoded
+  ///       text/*
+  ///
+  static Map<String, Map<String, Function>> decoders = {
```
anachlas: Could this be `_decoders`?

anachlas:
👍

lib/base/body_decoder.dart
joeconwaystk: I suppose it could be. I left encoders on Request public as well, thinking maybe someone could call them independently and wanted to keep them consistent... but decoders has a specific decode method that does the lookup. And I think it'd be better for encoders to have the same, and both of them will be private. i'll make that change

anachlas:
:+1:

PR #71, opened 2016-08-05
Jc/controllercaching

joeconwaystk:
This is simply optimizes controller request handler selection by caching it per isolate the first time a controller is instantiated.

### DIFF
--- a/lib/base/http_controller.dart
+++ b/lib/base/http_controller.dart
@@ -180,8 +180,17 @@ abstract class HTTPController extends RequestHandler {
   }
 
   Future<Response> _process() async {
-    var methodSymbol = _routeMethodSymbolForRequest(request);
-    var handlerParameters = _parametersForRequest(request, methodSymbol);
+    var key = _generateHandlerMethodKey(request.innerRequest.method, request.path.orderedVariableNames);
+    var cachedMethod = _methodCache[this.runtimeType][key];
+    if (cachedMethod == null) {
+      return new Response.notFound();
+    }
+
+    var methodSymbol = cachedMethod.methodSymbol;
+    var handlerParameters = cachedMethod
+        .orderedPathParameters
+        .map((param) => _convertParameterWithMirror(this.request.path.variables[param.name], param.typeMirror))
+        .toList();
 
     if (request.innerRequest.contentLength > 0) {
       if (_requestContentTypeIsSupported(request)) {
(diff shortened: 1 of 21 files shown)


lib/base/model_controller.dart
```diff
@@ -24,35 +24,35 @@ abstract class ModelController<T extends Model> extends HTTPController {
...
         var primaryKeyDesc = query.entity.attributes[query.entity.primaryKey];
         if (primaryKeyDesc.isAssignableWith(idValue)) {
           query[query.entity.primaryKey] = idValue;
         } else if (primaryKeyDesc.type == PropertyType.bigInteger || primaryKeyDesc.type == PropertyType.integer) {
           try {
             query[query.entity.primaryKey] = int.parse(idValue);
           } on FormatException {
-            var errorMessage = "Expected integer value for ModelController on ${query.entity}, but $idValue was not able to be parsed to an integer.";
+            var errorMessage = "Expected integer value for ModelController on ${query
+                .entity}, but $idValue was not able to be parsed to an integer.";
```
smswz: did this get autoformatted?

lib/base/http_controller.dart
```diff
@@ -246,6 +235,57 @@ abstract class HTTPController extends RequestHandler {
...
+    var methodMap = {};
+    var allDeclarations = reflect(this).type.declarations;
+    for (var key in allDeclarations.keys) {
+      var declaration = allDeclarations[key];
+      if (declaration is MethodMirror) {
+        var methodAttrs = declaration
+            .metadata
+            .firstWhere((attr) => attr.reflectee is HTTPMethod, orElse: () => null);
+
+        if (methodAttrs != null) {
```
smswz: how about:
```
if (methodAttrs == null) { continue; }
```
To prevent another level of indention?

lib/base/http_controller.dart
```diff
@@ -246,6 +235,57 @@ abstract class HTTPController extends RequestHandler {
...
+                return new _HTTPControllerCachedParameter()
+                  ..name = MirrorSystem.getName(pm.simpleName)
+                  ..typeMirror = pm.type;
+              });
+
+          var generatedKey = _generateHandlerMethodKey((methodAttrs.reflectee as HTTPMethod).method, params.map((p) => p.name).toList());
+          var cachedMethod = new _HTTPControllerCachedMethod()
+            ..methodSymbol = key
+            ..orderedPathParameters = params
+            ..optionalParameters = new Map.fromIterable(optionalParams, key: (p) => p.name, value: (p) => p);
```
smswz: why not store the optional parameters above as a map instead of in the object initialization?

lib/base/model_controller.dart
joeconwaystk: ugh, yes

lib/base/http_controller.dart
```diff
@@ -246,6 +235,57 @@ abstract class HTTPController extends RequestHandler {
...
   }
 
+  void _buildCachesIfNecessary() {
+    if (_methodCache.containsKey(this.runtimeType)) {
+      return;
+    }
+
+    var methodMap = {};
+    var allDeclarations = reflect(this).type.declarations;
+    for (var key in allDeclarations.keys) {
```
smswz: can you not do a `allDeclarations.forEach(` call?

lib/base/http_controller.dart
joeconwaystk: when I can, I prefer to use methods on Iterable for transforming. unfortunately, Map has no such methods. It is more readable than a for loop that aggregates a map. Unless I am misunderstanding what you mean

lib/base/http_controller.dart
joeconwaystk: Yes, changed that as that is better

lib/base/http_controller.dart
smswz: I meant have the transform to a map at the end of the variable declaration, so you have a nice `Map<String,_HTTPControllerCachedParameter> optionalParams` that you can just use like: `..optionalParameters = optionalParams`.

To give a consistent look between `params` and `optionalParams`. Initially, I thought you were going to do something with the optional params since its an iterable and not a map (like the list of params).

joeconwaystk: Got it. I can't think of a way to do that, because Map and Iterable don't get along well (other than that .fromIterable named constructor). Unless you mean just move the fromIterable up to underneath it

PR #73, opened 2016-08-11
Add annotations for query and header values

smswz:
Allows having both controller level variables and method variables. Going forward, all named argument parameters to routes must be annotated with `@HTTPQuery` or `@HTTPHeader` to be matched.


lib/base/http_controller.dart
```diff
@@ -120,20 +121,96 @@ abstract class HTTPController extends RequestHandler {
...
-    values.forEach((key, value) {
-      var desired = desiredParameters[key];
-      if (desired != null) {
-        if (value is List) {
-          returnValue[new Symbol(key)] = _convertParameterListWithMirror(value, desired.typeMirror);
-        } else {
-          returnValue[new Symbol(key)] = _convertParameterWithMirror(value, desired.typeMirror);
+  Map<String, dynamic> _headerParametersFromRequest(Request request) {
+    Map<String, dynamic> headerParams = {};
+    request.innerRequest.headers.forEach((k,v) => headerParams[k] = v);
```
joeconwaystk: Should we collect every parameter here? This is just copying the map of headers. We should really only collect the ones that we care about collecting for the request handler method.

lib/base/http_controller.dart
```diff
@@ -120,20 +121,96 @@ abstract class HTTPController extends RequestHandler {
...
+        value = _convertParameterListWithMirror(value, param.typeMirror);
+      } else if (value != null) {
+        value = _convertParameterWithMirror(value, param.typeMirror);
+      }
+
+      reflect(this).setField(sym, value);
+    });
+  }
+
+  String _checkForRequiredParameters(_HTTPControllerCachedMethod cachedMethod, Map<String, dynamic> queryParameters, Map<String, dynamic> headerParameters) {
```
joeconwaystk: Can these be converted into a chain of iterables on the keys? a map->where->map?

lib/base/http_controller.dart
```diff
@@ -120,20 +121,96 @@ abstract class HTTPController extends RequestHandler {
...
+        if (param.httpParameter is HTTPQuery && queryParameters[externalName] == null) {
+          missingQueries.add("'${externalName}'");
+        } else if (param.httpParameter is HTTPHeader && headerParameters[externalName.toLowerCase()] == null) {
+          missingHeaders.add("'${externalName}'");
         }
       }
     });
 
-    return returnValue;
+    if (missingQueries.isEmpty && missingHeaders.isEmpty) {
```
joeconwaystk: Can this be moved into a separate method that simply returns a String based on these two inputs?

lib/base/http_controller.dart
smswz: Unfortunately there isn't direct access to the map of values. There is only a `.forEach` method and a `[]` operator. This was done to mimic the behavior of the queryParameter version - not doing any parsing or matching, but simply getting the information from a location into a common format.

joeconwaystk:
Because it's in my head right now, continuing on our talk from the end of the day: it'd make sense to have a Request -> Request Handler Method mapper class/method. You'd send it the handler method info after finding a match (the cached stuff) and then the incoming request. It'd return a structure that has Symbol -> Value maps for query params and header params. Perhaps we even move path params in there as well? maybe all of the matching/method finding/caching moves there as well? If you can just pass a request into a collection of these things and either get out null (no handler method), or a structure that contains error information, or a structure that contains the maps AND symbol for the method to execute.

lib/auth/auth_code_controller.dart
```diff
@@ -24,12 +24,13 @@ class AuthCodeController extends HTTPController {
...
-      return new Response.badRequest();
-    }
-
-    var authCode = await authenticationServer.createAuthCode(username, password, client_id);
+  Future<Response> authorize({
+                             @HTTPQuery.required("client_id") String clientID,
+                             @HTTPQuery.required("username") String username,
+                             @HTTPQuery.required("password") String password,
+                             @HTTPQuery.optional("state") String state
+                             }) async {
```
joeconwaystk: I feel like the following looks better:

```
Future<Response> authorize({
  @HTTPQuery.required("client_id") String clientID,
  @HTTPQuery.required("username") String username,
  @HTTPQuery.required("password") String password,
  @HTTPQuery.optional("state") String state
}) async {
```

lib/auth/auth_controller.dart
```diff
@@ -27,30 +30,33 @@ class AuthController extends HTTPController {
   /// When grant_type is 'refresh', there must be a refresh_token value.
   /// When grant_type is 'authorization_code', there must be a authorization_code value.
   @httpPost
-  Future<Response> create({String grant_type, String username, String password, String refresh_token, String authorization_code}) async {
-    var authorizationHeader = request.innerRequest.headers[HttpHeaders.AUTHORIZATION]?.first;
-
-    var basicRecord = AuthorizationBasicParser.parse(authorizationHeader);
-    if (grant_type == "password") {
+  Future<Response> create({
```
joeconwaystk: See previous comment on alignment, same here

lib/base/resource_controller.dart
```diff
@@ -205,12 +205,19 @@ class ResourceController<ModelType extends Model> extends HTTPController {
...
   /// Executed after a list of objects has been fetched.
   ///
-  /// By defualt, returns [Response.ok] with the encoded list of founds objects (which may be the empty list).
+  /// By default, returns [Response.ok] with the encoded list of founds objects (which may be the empty list).
   Future<Response> didFindObjects(List<ModelType> objects) async {
     return new Response.ok(objects);
   }
 
-  @httpGet getObjects({int count: 0, int offset: 0, String pageBy: null, String pageAfter: null, String pagePrior: null, List<String> sortBy: null}) async {
+  @httpGet getObjects({
```
joeconwaystk: param align

test/base/controller_test.dart
```diff
@@ -315,24 +421,50 @@ class DateTimeController extends HTTPController {
...
 
+class HTTPParameterController extends HTTPController {
+  @HTTPHeader.optional("Location") String location;
+  @HTTPHeader.required("X-Request-id") String requestId;
+  @HTTPQuery.optional("number") int number;
+  @HTTPQuery.required("Shaqs") int numberOfShaqs;
+
+  @httpGet
+  Future<Response> get({
+                       @HTTPHeader.required("Cookie") String cookieBrand,
```
joeconwaystk: param align

lib/auth/auth_code_controller.dart
smswz: Were you able to get the editor to support that indentation? It's only formatted that way cause thats how the editor does it, there are no knobs to really configure the indent level of it

lib/base/parameter_matching.dart
```diff
@@ -0,0 +1,312 @@
...
+    var key = _generateHandlerMethodKey(req.innerRequest.method, req.path.orderedVariableNames);
+    var matchingMethod = _methodCache[controllerType][key];
+    if (matchingMethod == null) {
+      return null;
+    }
+
+    return new _HTTPMethodParameterTemplate._internal(con, req, matchingMethod.methodSymbol);
+  }
+
+  _HTTPMethodParameterTemplate._internal(HTTPController this._controller, Request this._request, Symbol this._methodSymbol);
```
joeconwaystk: You can omit the type info here

lib/base/parameter_matching.dart
```diff
@@ -0,0 +1,312 @@
...
+
+  static void _buildCachesIfNecessary(Type controllerType) {
+    if (_methodCache.containsKey(controllerType)) {
+      return;
+    }
+
+    var controllerLevelMap = {};
+    var methodMap = {};
+    var allDeclarations = reflectClass(controllerType).declarations;
+    allDeclarations.forEach((key, declaration) {
```
joeconwaystk: Can these get split up so that we can avoid the inner if/else? So:

```
allDeclarations.where((decl) => decl is VariableMirror).forEach((varMirror) { ... do property stuff ... });

allDeclarations.where((decl) => decl is MethodMirror).forEach((methodMirror) { .. do method stuff });
```

joeconwaystk: and note that in Dart 1.19, the analyzer will likely be able to infer the type of the args for the forEach, so dealer's choice on whether you type them or not

lib/base/parameter_matching.dart
```diff
@@ -0,0 +1,312 @@
...
+      if (declaration is VariableMirror) {
+        _HTTPParameter httpParameter = declaration.metadata.firstWhere((im) => im.reflectee is _HTTPParameter, orElse: () => null)?.reflectee;
+        if (httpParameter == null) { return; }
```
joeconwaystk: newline. This early exit format is prone to readability issues when someone deletes the newline underneath it

lib/base/parameter_matching.dart
```diff
@@ -0,0 +1,312 @@
+part of aqueduct;
+
+/// Parent class for annotations used for optional parameters in controller methods
+abstract class _HTTPParameter {
```
joeconwaystk: Can you verify that externalName and isRequired show up in the generated documentation for aqueduct API ref in both HTTPHEader and HTTPQuery? Run dartdoc on the package. Want to make sure that because _HTTPParameter is private, that those variables still show up appropriately.

smswz: They do show up

lib/base/parameter_matching.dart
```diff
@@ -0,0 +1,312 @@
...
+
+        if (methodAttrs == null) {
+          return;
+        }
+
+        List<_HTTPControllerCachedParameter> params = (declaration as MethodMirror)
+            .parameters
+            .where((pm) => !pm.isOptional)
+            .map((pm) {
+          var name = MirrorSystem.getName(pm.simpleName);
```
joeconwaystk: IntelliJ does this and it pisses me off, but slide the body of map in a few indents

lib/base/parameter_matching.dart
```diff
@@ -0,0 +1,312 @@
...
+            ..name = name
+            ..typeMirror = pm.type;
+        })
+            .toList();
+
+        Iterable<_HTTPControllerCachedParameter> optionalParams = (declaration as MethodMirror)
+            .parameters
+            .where((pm) => pm.metadata.any((im) => im.reflectee is _HTTPParameter))
+            .map((pm) {
+          _HTTPParameter httpParameter = pm.metadata.firstWhere((im) => im.reflectee is _HTTPParameter).reflectee;
```
joeconwaystk: indent

lib/base/parameter_matching.dart
```diff
@@ -0,0 +1,312 @@
...
+
+    _methodCache[controllerType] = methodMap;
+    _controllerLevelParameters[controllerType] = controllerLevelMap;
+  }
+
+  static String _generateHandlerMethodKey(String httpMethod, List<String> params) {
+    return "${httpMethod.toLowerCase()}-" + params.map((pathParam) => pathParam).join("-");
+  }
+
+  _HTTPMethodParameterValues parseRequest() {
```
joeconwaystk: This one is kind of confusing me. So these symbolicate methods build up and return a map of values, but if there are errors, they set state in the instance, and then a combo of those maps and that state are used to create a new instance? Is there anyway to make those symbolicate methods avoid side effects?

lib/base/parameter_matching.dart
joeconwaystk: Ah, I didn't even recognize that it was a map. OK. agreed

joeconwaystk: What about a method that maps a VariableMirror -> _HTTPCOntrollerCachedParameter and a MethodMirror -> _HTTPControllerCachedMethod? At least that way we could split up this method a bit.

```
if (decl is VariableMirror) {
  var param = _cachedParameterFrom(decl);
  if (param != null) {
    controllerLevelMap[key] = param;
  }
} else if ...
```

lib/base/parameter_matching.dart
smswz: I can have them store the maps as instance state as well, unfortunately cant return two things from one function, which is ideally what I'd want to do.

The symbolicating of values and the error checking is done inline because the traversal code is exactly the same, and wanted to avoid writing the same traversal code twice.

joeconwaystk: Agreed on avoiding the double traversal. Can you instantiate the _HTTPMethodParameterValues early and pass it to each of these functions to avoid that? A tuple would be valuable here, but alas

joeconwaystk:
Looks like _methodCache and _controllerLevelParameters are left on HTTPController

PR #75, opened 2016-08-18
Jc/combinewildfire

joeconwaystk:
Moving Wildfire into Aqueduct - fixing up a few scripts.


bin/create.dart
```diff
@@ -0,0 +1,236 @@
...
+    ".travis.yml"
+  ];
+
+  var lastComponent = entity.uri.pathSegments.last;
+  if (lastComponent.length == 0) {
+    lastComponent = entity.uri.pathSegments[entity.uri.pathSegments.length - 2];
+  }
+
+  if (lastComponent.startsWith(".")) {
+    if (!hiddenFilesToKeep.contains(lastComponent)) {
```
smswz: move into one if statement

bin/create.dart
```diff
@@ -0,0 +1,236 @@
...
+
+  new File(projectDirectory.path + "/pubspec.yaml").deleteSync();
+  new File(projectDirectory.path + "/.packages").deleteSync();
+
+  var lastLibIndex = resolvedURL.lastIndexOf("/lib");
+  return resolvedURL.substring(0, lastLibIndex);
+}
+
+bool shouldIncludeItem(FileSystemEntity entity) {
+  var ignoreFiles = [
```
smswz: should these constants be moved out of the function so they aren't created on each run?

bin/generate_client_id.dart
```diff
@@ -0,0 +1,15 @@
+import 'dart:io';
+import 'package:aqueduct/aqueduct.dart';
+
+void main(List<String> args) {
+  if (args.length != 2) {
+    print("Usage: pub global run aqueduct:generate_client_id client_id client_secret");
```
smswz: does the user provide the client_id and client_secret?

example/templates/default/restart.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+git fetch
+git checkout master
+git reset --hard
+git pull
+
+/usr/lib/dart/bin/pub get
+
+pkill dart
```
smswz: wont this kill all dart processes?

joeconwaystk: It does, but this is a server script so there shouldn't be anything else running.

joeconwaystk: I can't think of a way other than recording the pid somewhere on run and then using that, but even that, I don't know if that guarantees anything...

bin/generate_client_id.dart
joeconwaystk: Yes, this is just a handy tool for generating a client id/secret that has been hashed so you can insert it into a db

bin/create.dart
joeconwaystk: Potentially, but I'm not all the concerned about it now as this is just a one time quick script

example/templates/default/restart.sh
smswz: I know it kills the dart analyzer for intelliJ. We could add a dotfile recording the PID on start if we're worried

joeconwaystk: Is there anything to worry about in terms of those two things getting out of sync? I guess one could manually kill the process at that point.

smswz: I guess it could just be a best effort, if the pid is still running when trying to rerun, a bind socket will fail and hopefully the user would know what to do. idk if thats an entirely obvious solution though

bin/create.dart
```diff
@@ -219,7 +219,8 @@ String generatingPubspec(String aqueductDependencyString) {
 }
 
 bool isSnakeCase(String string) {
-  return string.toLowerCase() == string;
+  var expr = new RegExp("^[a-z][a-z0-9_]*\$");
```
smswz: Is the first bracket supposed to be `[A-Z]`?

joeconwaystk: negative, must be lowercase. It can't start with a number or _, must be a lowercase letter

smswz: ah, yea, was thinking camel case

smswz:
any other changes?

joeconwaystk:
Good to go - we'll eventually improve those scripts for startup, but these were clearly a step in the right direction

PR #76, opened 2016-08-23
Responses auto doc generation

smswz:


smswz:
I think I covered everything and should be good to go @joeconwaystk

lib/auth/auth_code_controller.dart
```diff
@@ -52,4 +53,23 @@ class AuthCodeController extends HTTPController {
...
+  List<APIResponse> documentResponsesForOperation(APIOperation operation) {
+    if (operation.id == APIOperation.idForMethod(this, #authorize)) {
+      return [
+        new APIResponse()
+          ..statusCode = HttpStatus.MOVED_TEMPORARILY
+          ..description = "Successfully issued an authorization code.",
+        new APIResponse()
+          ..statusCode = HttpStatus.BAD_REQUEST
+          ..description = "Missing one or more of: 'client_id', 'username', 'password'.",
+        new APIResponse()
```
joeconwaystk: If default is something went wrong, should it also indicate a 500 status code and be appended to any list of APIResponses that don't include it instead of having to specifically include it? It is necessary as part of the spec?

lib/auth/auth_controller.dart
```diff
@@ -74,4 +74,38 @@ class AuthController extends HTTPController {
...
+  List<APIResponse> documentResponsesForOperation(APIOperation operation) {
+    if (operation.id == APIOperation.idForMethod(this, #create)) {
+      return [
+        new APIResponse()
+          ..statusCode = HttpStatus.OK
+          ..description = "Successfully exchanged credentials for credentials"
+          ..schema = (new APISchemaObject()
+            ..type = APISchemaObjectTypeObject
+            ..properties = {
+              "access_token" : new APISchemaObject()..type = APISchemaObjectTypeString,
```
joeconwaystk: can we get some named constructors for the common types so we don't have to specify it by setting a property? 
```
new APISchemaObject.string()
```

lib/base/resource_controller.dart
```diff
@@ -264,6 +264,89 @@ class ResourceController<ModelType extends Model> extends HTTPController {
...
   }
 
+  @override
+  List<APIResponse> documentResponsesForOperation(APIOperation operation) {
+    if(operation.id == APIOperation.idForMethod(this, #getObject)) {
+      return [
+        new APIResponse()
+          ..statusCode = HttpStatus.OK
+          ..description = ""
+          ..schema = ModelContext.defaultContext.entityForType(ModelType).documentedResponseSchema,
```
joeconwaystk: This should used the passed in context, which defaults to defaultContext.

lib/base/resource_controller.dart
```diff
@@ -264,6 +264,89 @@ class ResourceController<ModelType extends Model> extends HTTPController {
...
+              "error" : new APISchemaObject()..type = APISchemaObjectTypeString
+            }
+          ),
+      ];
+    } else if (operation.id == APIOperation.idForMethod(this, #createObject)) {
+      return [
+        new APIResponse()
+          ..statusCode = HttpStatus.OK
+          ..description = ""
+          ..schema = ModelContext.defaultContext
```
joeconwaystk: use passed in context

lib/db/data_model.dart
```diff
@@ -69,21 +69,36 @@ class DataModel {
...
+    Map<Symbol, DeclarationMirror> persistentDeclarations = entity.persistentInstanceTypeMirror.declarations;
+    entity.instanceTypeMirror.declarations.values
+      .where((declMir) => declMir is VariableMirror && !declMir.isStatic)
+      .where((declMir) => !declMir.metadata.any((im) => im.type.isSubtypeOf(reflectType(Relationship))))
+      .forEach((declMir) {
+        var key = MirrorSystem.getName(declMir.simpleName);
+        var persistentMirror = persistentDeclarations[declMir.simpleName];
+        var isMappable = declMir.metadata.any((im) => im.type.isSubtypeOf(reflectType(Mappable)));
+
+        if (persistentMirror != null || isMappable) {
```
joeconwaystk: Shouldn't this be an &&?

lib/db/data_model.dart
```diff
@@ -69,21 +69,36 @@ class DataModel {
...
+          map[key] = _attributeFromVariableMirror(entity, declMir, persistentMirror);
         }
-      }
     });
+
+    persistentDeclarations.values
+      .where((declMir) => declMir is VariableMirror && !declMir.isStatic)
+      .where((declMir) => !declMir.metadata.any((im) => im.type.isSubtypeOf(reflectType(Relationship))))
+      .where((declMir) => !map.containsKey(MirrorSystem.getName(declMir.simpleName)))
```
joeconwaystk: I guess you are doubling back here and avoiding anything that was already put into the map because it's also in the instanceTypeMirror, but shouldn't everything in the persistentInstanceTypeMirror also be in the instanceTypeMirror?

lib/db/data_model.dart
```diff
@@ -69,21 +69,36 @@ class DataModel {
...
+        map[key] = _attributeFromVariableMirror(entity, null, declMir);
+      });
+
     return map;
   }
 
-  AttributeDescription _attributeFromVariableMirror(ModelEntity entity, VariableMirror mirror) {
-    Attributes metadataAttrs = mirror.metadata
-        .firstWhere((im) => im.type.isSubtypeOf(reflectType(Attributes)), orElse: () => null)
+  AttributeDescription _attributeFromVariableMirror(ModelEntity entity, VariableMirror instanceMirror, VariableMirror persistentMirror) {
```
joeconwaystk: make both instanceMirror and persistentMirror optional args, that way it is at least indicative. But I think you can better determine of the VariableMirror is part of the persistent/instance type by just checking its owner value against the persistentInstanceTypeMirror/instanceTypeMirror, and that way you don't have to separate them

lib/db/model.dart
```diff
@@ -109,17 +109,14 @@ class Model<T> implements Serializable {
...
-        if(decl != null && _declarationMirrorIsMappableOnInput(decl)) {
-          reflect(this).setField(decl.simpleName, v);
-        } else {
-          throw new QueryException(400, "Key $k does not exist for ${MirrorSystem.getName(reflect(this).type.simpleName)}", -1);
-        }
-      } else {
+      if (property != null && !(property is AttributeDescription && property.isTransient)) {
         dynamicBacking[k] = _valueDecoder(property, v);
+      } else if (decl != null && _declarationMirrorIsMappableOnInput(decl)) {
+        reflect(this).setField(decl.simpleName, v);
```
joeconwaystk: Shouldn't this also go thru the _valueDecoder?

lib/db/model_entity.dart
```diff
@@ -28,6 +28,84 @@ class ModelEntity {
   /// The [DataModel] this instance belongs to.
   final DataModel dataModel;
 
+  /// Schema of the model as returned from a request
+  APISchemaObject get documentedResponseSchema {
```
joeconwaystk: Since this will only ever get called during documentation, it's not particularly speed sensitive (nor will be called often) - is there another reason for caching it that I'm not seeing?

lib/db/model_entity.dart
```diff
@@ -28,6 +28,84 @@ class ModelEntity {
...
+        ..title = MirrorSystem.getName(instanceTypeMirror.simpleName)
+        ..type = APISchemaObjectTypeObject
+        ..properties = _propertiesForEntity(this);
+    }
+
+    return _documentedResponseSchema;
+  }
+  APISchemaObject _documentedResponseSchema;
+
+  Map<String, APISchemaObject> _propertiesForEntity(ModelEntity me, {bool shallow: false}) {
```
joeconwaystk: _schemaForEntity?

joeconwaystk: Is the intent here to either get a typical SchemaObject or get the entire tree and have the developer prune it themselves later?

lib/db/model_entity.dart
```diff
@@ -28,6 +28,84 @@ class ModelEntity {
...
+    Map<String, APISchemaObject> schemaProperties = {};
+    me.attributes.values
+      .where((attribute) => attribute.isIncludedInDefaultResultSet)
+      .where((attribute) => !shallow || attribute.isPrimaryKey)
```
joeconwaystk: So basically, if it's shallow, only grab the primary key? Seems like we don't really need to iterate over the whole thing to make that determination, just pluck it and go.

lib/db/model_entity_property.dart
```diff
@@ -100,8 +108,9 @@ abstract class PropertyDescription {
 /// Each non-relationship property [Model] object persists is described by an instance of [AttributeDescription]. This class
 /// adds two properties to [PropertyDescription] that are only valid for non-relationship types, [isPrimaryKey] and [defaultValue].
 class AttributeDescription extends PropertyDescription {
-  AttributeDescription(ModelEntity entity, String name, PropertyType type, {bool primaryKey: false, String defaultValue: null, bool unique: false, bool indexed: false, bool nullable: false, bool includedInDefaultResultSet: true, bool autoincrement: false}) :
+  AttributeDescription(ModelEntity entity, String name, PropertyType type, {bool transient: false, bool primaryKey: false, String defaultValue: null, bool unique: false, bool indexed: false, bool nullable: false, bool includedInDefaultResultSet: true, bool autoincrement: false}) :
```
joeconwaystk: Hm. I'm not sure I want the ability to define a transient Attribute. I think by default, an Attribute or anything in the persistentEntityType must be persistent. Is this setup for another purpose?

joeconwaystk: Oh sorry, I misread this, this is the AttribureDescription, not Attributes. Will delete if I can

lib/db/schema_generator.dart
```diff
@@ -116,6 +116,9 @@ class SchemaColumn {
       case PropertyType.boolean: return "boolean";
       case PropertyType.datetime: return "datetime";
       case PropertyType.string: return "string";
+      case PropertyType.transientList:
```
joeconwaystk: Let's avoid fall thru  and be explicit :)

joeconwaystk:
One thing I am seeing missing overall is RequestHandlers like Authenticator should additionally their possible responses to the options. This is also why the structure of Documentable is set up in that way. Basically, Authenticator will override the documentResponses method to first call its documentableChild's version, and then after collecting those, will add its own possible responses to that list. If there is a conflict, I'm not sure how that is resolved, will have to review that more closely...

lib/db/data_model.dart
smswz: Some things were not being in both when running the tests, causing the tests the fail. The primaryKey property was in the persistent entity but not the instance type, causing the primaryKey to never be added to the map and failing a whole bunch of tests.

lib/db/model_entity.dart
smswz: this method is for filling out the `properties` field in a schema object, which recursively gets all properties all the way down the tree.

I'm not sure what you are asking... why would the developer prune the tree?

lib/db/model_entity.dart
smswz: I guess its not speed sensitive, and each model entity should only be called deeply once, probably doesn't need to be cached, yea.

lib/db/data_model.dart
smswz: No, want to get all persistent properties and all transient properties

lib/auth/auth_code_controller.dart
smswz: It's not required, but its intended for 500s as well as other error codes that someone forgets to add. I'll move it to the default responses generated in the parent `http_controller` class

joeconwaystk: This is now included in the base class so no need to add it here.

lib/base/documentable.dart
```diff
@@ -231,6 +232,19 @@ class APIOperation {
   APIRequestBody requestBody;
   List<APIResponse> responses = [];
 
+  static String idForMethod(Object classInstance, Symbol methodSymbol) {
+    return "${MirrorSystem.getName(reflect(classInstance).type.simpleName)}.${MirrorSystem.getName(methodSymbol)}";
+  }
+
+  static Symbol symbolForId(String operationId, Object classInstance) {
```
joeconwaystk: Use capital ID to remain consistent. Swift and Dart both

lib/base/parameter_matching.dart
```diff
@@ -72,6 +76,21 @@ class _HTTPControllerCache {
 
   Map<String, _HTTPControllerCachedMethod> methodCache = {};
   Map<Symbol, _HTTPControllerCachedParameter> propertyCache = {};
+  bool _hasControllerRequiredParameter = false;
+
+  bool hasRequiredParametersForMethod(MethodMirror mm) {
+    if (_hasControllerRequiredParameter) {
```
joeconwaystk: Not sure how this value helps? Aren't we just finding out the same thing in the next code block?

test/db/data_model_test.dart
```diff
@@ -188,13 +188,50 @@ void main() {
...
+    });
+  });
 }
 
 class User extends Model<_User> implements _User {}
 class _User {
   @primaryKey
   int id;
 
+  @mappable
```
joeconwaystk: This is poorly documented, but mappable is only valid for properties in the instance type, aka transient properties.  All persistent properties are mappable and so this has no effect. Specifying something like mappableInput or mappableOutput also would not have an impact on a persistent property. This should probably throw a hard exception at runtime.

test/db/data_model_test.dart
```diff
@@ -188,13 +188,50 @@ void main() {
...
+  String stringId;
```
joeconwaystk: always use 'ID' not 'Id'

lib/base/parameter_matching.dart
smswz: No, this is to test for required parameters on the controller, the next block only checks for required parameters from the methods optional parameters. Since you can't determine the controller's parameters strictly from the method, I cached this so you don't have to iterate through all a classes declarations on calling this method

joeconwaystk: Got it. Makes sense

PR #77, opened 2016-09-01
Add schema for default error

smswz:


joeconwaystk:
Can we get a test to ensure that a null schema object produces valid document object? In general, a bug should always yield a fix/test pair

PR #78, opened 2016-09-08
Make it so randomly null things don't crash the isolate.

anachlas:


joeconwaystk:
I updated this from the changes I made on the main branch as a patch for this branch. You have to merge now. Check out the merge into master as well.

anachlas:
👍

PR #79, opened 2016-09-09
Jc/fixcatch

joeconwaystk:
This PR addresses the issue that can cause two isolate crashers: trying to send a 500 response after a response has already been sent and logging after closing a response doesn't crash in strange circumstances.


anachlas:
👍

PR #81, opened 2016-09-19
Jc/refactorquery

joeconwaystk:
(Note that this PR merges into 1.0 branch, which is WIP and need not be fully functional yet.)

Checking in this PR now, even though more testing needs to be done (and I know there is one failing test). These issues will be noted in pivotal. However, this was a really far reaching refactor stemming from adding strong mode, so this is a reasonable stopping point to break up the subsequent work into more manageable tasks.

This PR also points at a temporary holding place for the postgresql driver. This will be replaced in a subsequent PR when the new postgresql driver is ready.

There are a lot of big changes.

1. Updated to use strong mode.
2. Removed ModelQuery, significantly improved interface for handling relationships and joins.
3. Altered the behavior of transient properties.
4. Refactored some code related to building data models.
5. Renamed a lot of properties to make them easier to understand.
6. Started building significant guide documentation in docs/


doc/db/executing_queries.md
```diff
@@ -0,0 +1,426 @@
...
+user.name; // "Bob"
+user.email; // "bob@stablekernel.com"
+user.id; // 1
+```
+
+Since most model objects have an auto-incrementing primary key, it is important to get the returned model object from an insert execution, as it will contain the generated primary key value.
+
+Note that only the data explicitly set in `values` (or `valueMap`) is sent to the database. Unset values are *not* sent as `null` and entirely omitted from the insert command.
+
+If an insert query fails because of a conflict - a unique constraint is violated - Aqueduct will automatically generate an `HTTPResponseException` with a 409 status code. This default behavior means that you don't have to check for the specific error of unique constraint violation and Aqueduct will abandon execution and return a 409 'Conflict' response to the calling client.
```
rosshambrick: I agree with @smswz. The controller should map database errors to HTTP errors.

doc/db/executing_queries.md
```diff
@@ -0,0 +1,426 @@
...
+A delete `Query` will delete rows from a database for the entity specified by its type argument. Like update and fetch queries, you may specify a row or rows using `matchOn` and `predicate` properties of the `Query`. The result of a delete operation will be a `Future` with the number of rows deleted.
+
+```dart
+var query = new Query<User>()
+  ..matchOn.id = 1;
+
+int usersDeleted = await query.delete();
+```
+
+Also like update queries, delete queries have a safety feature that prevents you from accidentally deleting every row in a table. If you try to execute a `Query` to do an update and no `predicate` or `matchOn` property is defined, the default behavior of `Query` will throw an exception prior to carrying out the request. If you explicitly want to delete every instance of some entity (that is, every row of a table), you must set the `Query`'s `confirmQueryModifiesAllInstancesOnDeleteOrUpdate` to `true` prior to execution.
```
smswz: Is there any way to reduce the length of the paragraph since it's almost exactly the same 2x? Maybe add a note from one to another?

doc/db/executing_queries.md
```diff
@@ -0,0 +1,426 @@
...
+
+user.id == 2;
+user.tasks.every((Task t) =>
+  t.id == 1 &&
+  t.user.id == 2 &&
+  t.text is String
+) == true;
+```
+
+As shown, you may still apply matcher expressions to the primary entity of the query. You may also apply matchers to the related objects. In the case of hasOne relationships, this doesn't make much sense - one you've included the only possible related object, filtering do anything useful. Thus, when fetching hasOne properties, you need only set the relationships `includeInResultSet` property.
```
smswz: typo: `filtering do anything useful`

doc/db/executing_queries.md
```diff
@@ -0,0 +1,426 @@
...
+  ..nestedResultProperties[Task] = ["id", "text"]
+  ..matchOn.id = 1
+  ..matchOn.tasks.includeInResultSet = true;
+```
+
+Note that a query will always add the primary key of a nested object if it is omitted in `nestedResultProperties`.
+
+### Exceptions and Errors
+
+An error encountered in preparing or executing a query will throw a `QueryException`. A `QueryException` extends `HTTPResponseException`, and therefore has a suggested HTTP status code. For common scenarios - like a unique violation generating an exception with suggested status code of `409` - Aqueduct will return a reasonable status code to the requesting HTTP client. Therefore, you do not have to catch query executions unless you wish to override the returned status code.
```
smswz: Can exceptions implement interfaces? Or mixins? A `QueryException` does not seem like a logical child class of `HTTPResponseException`, but it does make sense to have built-in shortcut behavior for a failed query.

doc/db/modeling_data.md
```diff
@@ -0,0 +1,258 @@
...
+  new User()
+    ..id = 1
+    ..name = "Bob",
+  new User()
+    ..id = 2
+    ..name = "Fred"
+];
+```
+
+instance types may also define properties and methods on top of those it implements from its persistent type. Because these properties and methods are not part of the persistent type, they are *transient* - that is, they are not stored in the database. Any method or property defined on an instance type is ignored when used in a `Query`. This is in contrast to a persistent type, where every property explicitly maps to a database column.
```
smswz: Capitalize `instance`

lib/base/http_controller.dart
```diff
@@ -174,7 +174,10 @@ abstract class HTTPController extends RequestHandler {
 
     var classUnit = fileUnit.declarations
         .where((u) => u is ClassDeclaration)
-        .firstWhere((ClassDeclaration u) => u.name.token.lexeme == MirrorSystem.getName(reflectedType.simpleName));
+        .firstWhere((CompilationUnitMember u) {
+          ClassDeclaration classDecl = u;
```
smswz: why not do a map to `ClassDeclaration` before the `.firstWhere` like you did above for `MethodMirror`?

lib/base/request_handler.dart
```diff
@@ -18,7 +20,12 @@ class RequestHandler extends Object with APIDocumentable {
   APIDocumentable get documentableChild => nextHandler;
 
   Logger get logger => new Logger("aqueduct");
-  CORSPolicy policy = new CORSPolicy();
+
+  CORSPolicy get policy => _policy;
```
smswz: why add the internal variable + getter and setter?

doc/db/executing_queries.md
JesseBlack82: I think the mapping from QueryExceptions -> HttpResponseExceptions at the controller level (suggested way above) could preserve the shortcut behavior without using inheratance

example/templates/default/lib/src/model/token.dart
```diff
@@ -1,28 +1,26 @@
 part of wildfire;
 
 class AuthCode extends Model<_AuthCode> implements _AuthCode {}
-class _AuthCode implements TokenExchangable {
+class _AuthCode implements TokenExchangable<Token> {
   @primaryKey
   int id;
 
-  @Attributes(indexed: true)
+  @AttributeHint(indexed: true)
```
JesseBlack82: Why did you change `@Attributes` to `@AttributeHint`?

example/templates/default/lib/src/model/user.dart
```diff
@@ -1,8 +1,7 @@
 part of wildfire;
 
-class UserQuery extends ModelQuery<User> implements _User {}
 class User extends Model<_User> implements _User, Authenticatable {
-  @mappableInput
+  @transientInputAttribute
```
JesseBlack82: I like this name change

example/templates/default/lib/src/utilities/auth_delegate.dart
```diff
@@ -27,51 +27,53 @@ class WildfireAuthenticationDelegate implements AuthenticationServerDelegate<Use
...
+      ..matchOn.email = username
       ..resultProperties= ["email", "hashedPassword", "salt", "id"];
 
     return await userQ.fetchOne();
   }
 
-  Future<User> authenticatableForID(AuthenticationServer server, int id) async {
-    var userQ = new UserQuery()
-      ..id = id
+  Future<User> authenticatableForID(AuthenticationServer server, dynamic id) async {
```
JesseBlack82: Why did you have to go with `dynamic`?

lib/base/documentable.dart
```diff
@@ -281,7 +281,7 @@ class APIResponse {
 
     return {
       "description" : description,
-      "schema" : schema.asMap(),
+      "schema" : schema?.asMap(),
       "headers" : mappedHeaders
```
JesseBlack82: Will `headers` be returned in the `Map` when `schema` is nil?

lib/db/model_entity.dart
```diff
@@ -109,27 +110,33 @@ class ModelEntity {
...
+
   /// Creates an instance of this entity from a list of [MappingElement]s.
   ///
   /// This method is used by a [ModelContext] to instantiate entities from a row
   /// returned from a database. It will initialize all column values, including belongsTo
   /// relationships. It will not populate data from hasMany or hasOne relationships
   /// that were populated in a join query, as this is the responsibility of the context.
   Model instanceFromMappingElements(List<MappingElement> elements) {
-    Model instance = instanceTypeMirror.newInstance(new Symbol(""), []).reflectee;
+    Model instance = newInstance();
```
JesseBlack82: That cleaned up nicely

lib/db/model_entity.dart
```diff
@@ -154,21 +161,22 @@ class ModelEntity {
     me.attributes.values
         .where((attribute) => attribute.isIncludedInDefaultResultSet)
         .forEach((attribute) {
-      schemaProperties[attribute.name] = new APISchemaObject()
-        ..title = attribute.name
-        ..type = _schemaObjectTypeForPropertyType(attribute.type)
-        ..format = _schemaObjectFormatForPropertyType(attribute.type);
-    });
+      print ("$attribute");
```
JesseBlack82: Do you want to leave this print statement?

lib/db/data_model_builder.dart
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  Map<String, RelationshipDescription> relationshipMapForEntity(ModelEntity entity) {
+    Map<String, RelationshipDescription> map = {};
+
+    entity.persistentType.declarations.forEach((sym, declMir) {
+      if (declMir is VariableMirror && !declMir.isStatic) {
+        var key = MirrorSystem.getName(sym);
+
+        if (_doesVariableMirrorRepresentRelationship(declMir)) {
```
smswz: combine this `if` statement with one above and create `key` in the resulting `if` statement

doc/db/executing_queries.md
joeconwaystk: OK, I'll add an item in pivotal to investigate the best way to handle this. The question is basically whether all RequestHandlers should have this (I think so) or just HTTP controllers.

doc/db/executing_queries.md
joeconwaystk: OK, resolved

doc/db/executing_queries.md
joeconwaystk: Resolved

example/templates/default/lib/src/model/token.dart
joeconwaystk: When writing the docs, an entity has 'attributes'. It was getting really confusing to refer to the 'attributes of the attributes'.

joeconwaystk: That being said, this is not a name I am partial to and so any feedback here would be great

example/templates/default/lib/src/utilities/auth_delegate.dart
joeconwaystk: So an authenticatable doesn't necessarily have to have an integer primary key. And there isn't a way - at least I didn't find a way - to be able to be more specific about the type the represents the id. And now the analyzer complains about it if it isn't the matching type in the overridden method

lib/base/documentable.dart
joeconwaystk: I suppose so, should they not?

lib/db/model_entity.dart
joeconwaystk: good eye

lib/db/data_model_builder.dart
joeconwaystk: OK, made that change as well.

doc/db/modeling_data.md
joeconwaystk: OK

lib/base/documentable.dart
JesseBlack82: I am just suffering trauma from Dictionary literals in iOS. It wouldn't include the final pair of values in this exampleI am just suffering trauma from Dictionary literals in iOS. It wouldn't include the final pair of values in this example

test/db/postgresql/fetch_test.dart
```diff
@@ -284,214 +250,17 @@ void main() {
...
-    <|x                  |
-    <|- - - x            |
-     |< - - - x          |
-    <|- - - - - - - - - -|x
-     |        < - - - - x|
-     |          < - - - -|x
-    <|- - - - - - - - - -| nil
-     |          < - - - -| nil
-     ---------------------
-     */
```
JesseBlack82: You will be missed, ASCI chart

test/db/postgresql/paging_test.dart
```diff
@@ -0,0 +1,294 @@
...
+    <|x                  |
+    <|- - - x            |
+     |< - - - x          |
+    <|- - - - - - - - - -|x
+     |        < - - - - x|
+     |          < - - - -|x
+    <|- - - - - - - - - -| nil
+     |          < - - - -| nil
+     ---------------------
+     */
```
JesseBlack82: phew

example/templates/default/lib/src/model/token.dart
JesseBlack82: How about `ColumnAttribute`?

JesseBlack82:
Can we turn the travis integration back on for this branch? I see 4 commits went to CI

test/db/postgresql/paging_test.dart
joeconwaystk: This is honestly the thing I am most proud of in the entire library

joeconwaystk:
added travis integration for PRs into 1.0

example/templates/default/lib/src/model/token.dart
joeconwaystk: Hm.... yes. Yes I like that. Because it is specific tot he underlying column

lib/db/ordered_set.dart
```diff
@@ -0,0 +1,49 @@
+part of aqueduct;
+
+class OrderedSet<InstanceType extends Model> extends Object with ListMixin<InstanceType>, _QueryMatchableExtension implements QueryMatchable, Serializable  {
```
joeconwaystk: There is a SetMixin (that's how everything gets implemented by only implementing a few primitive methods, just like how NSArray works). I don't know how well it would work when combining or how much value it might add. Distinctness should be handled by the underlying DB layer (and it is)

lib/db/persistent_store_query.dart
```diff
@@ -1,59 +1,46 @@
 part of aqueduct;
 
+enum JoinType {
+  leftOuter
```
joeconwaystk: Yeah, the idea is that maybe more types will be supported, although I don't really see their usefulness at this point. But that's there just in case

lib/db/postgresql/postgresql_persistent_store.dart
```diff
@@ -122,7 +122,7 @@ class PostgreSQLPersistentStore extends PersistentStore {
 
   Future<List<MappingElement>> executeInsertQuery(PersistentStoreQuery q) async {
     var queryStringBuffer = new StringBuffer();
-    queryStringBuffer.write("insert into ${q.entity.tableName} ");
+    queryStringBuffer.write("insert into ${q.rootEntity.tableName} ");
```
joeconwaystk: Yes, have made that change for next commit

doc/db/executing_queries.md
rosshambrick: Could the router handle exceptions at the controller level and have auto-mappings for the obvious ones?  404, 409, etc

example/templates/default/lib/src/utilities/auth_delegate.dart
rosshambrick: Can it be a generic type?

doc/db/executing_queries.md
```diff
@@ -0,0 +1,426 @@
+## `Query` Instances Send Commands to a Database
+
+To send commands to a database - whether to fetch, insert, delete or update objects - you will create, configure and execute instances of `Query`. A `Query` allows you to create database operations in Dart code in the domain of your application, as opposed to writing SQL. `Query`s are executed against a `ModelContext`. By default, every `Query` uses `ModelContext.defaultContext`, but this can be changed during the instantiation of a `Query`.
+
+(Note that raw SQL may be sent to a database in an Aqueduct application using instances of `PersistentStore` directly. Queries are meant for data operations, not database management.)
```
rosshambrick: And possibly query tuning and performance optimization

joeconwaystk: 👍 added

doc/db/executing_queries.md
joeconwaystk: Once the router passes the request on, it is absolved of all responsibility. That is, the router isn't a parent to the controller. The router and controller both inherit from RequestHandler - which is gonna get renamed - but the logic to map QueryException -> an HTTP response will go there I believe

joeconwaystk: And for 404 for a GET /resource/:id is auto-managed in ResourceController<T>, but that's because that controller is specific enough to map a URI to a database entity

example/templates/default/lib/src/utilities/auth_delegate.dart
joeconwaystk: Yes, but in this case it wouldn't help - the generic type to this class is User, which determines the return value of this method. The type of this parameter is a property on User, which could be any property. There's not a lot the language can do at 'compile-time ' to enforce that the type of this parameter is the type of the right property of the generic User type

doc/db/executing_queries.md
```diff
@@ -0,0 +1,426 @@
...
+
+Instances of `Query` are generic. The type parameter indicates the entity you are fetching, inserting, deleting or updating. The type parameter must be a subclass of `Model` for which a fully formed `ModelEntity` exists. When a `Query` is created, it sets its `entity` property to `ModelEntity` that represents its type parameter:
+
+```dart
+var query = new Query<User>();
+
+query.entity.instanceType; // -> ClassMirror on 'User'
+```
+
+The `DataModel` that a `Query`'s `entity` must belongs to same `ModelContext` as the `Query`'s `context`.
```
rosshambrick: Something's missing from this sentence, but I'm not sure what.  All I know is that I'm confused.

Has `DataModel` been defined yet?  What is it?

It's also very confusing with multiple of the same type of terms.  DataModel, Entity, ModelContext, Context

doc/db/executing_queries.md
```diff
@@ -0,0 +1,426 @@
...
+
+A `Query` has four basic execution methods: `fetch`, `update`, `insert`, `delete`. These methods will perform a database operation based on the information configured in the `Query`.
```
rosshambrick: Why not go 4 for 4 on the SQL terminology?  `fetch` -> `select`

lib/db/ordered_set.dart
smswz: I was more interested because it is called a set, but doesn't necessarily implement set semantics

doc/db/executing_queries.md
joeconwaystk: I'm just going to axe that sentence honestly

lib/db/data_model_builder.dart
smswz: I dont think this change was made

joeconwaystk: This is what I see:

```
Map<String, RelationshipDescription> relationshipMapForEntity(ModelEntity entity) {
    Map<String, RelationshipDescription> map = {};

    entity.persistentType.declarations.forEach((sym, declMir) {
      if (declMir is VariableMirror
      && !declMir.isStatic
      && _doesVariableMirrorRepresentRelationship(declMir)) {
        var key = MirrorSystem.getName(sym);
        map[key] = relationshipFromVariableMirror(entity, declMir);
      }
    });

    return map;
  }
```

smswz: oh yep, there it is, this new review thing makes it impossible to find the correct revision

smswz:
🆗  with me

doc/db/executing_queries.md
rosshambrick: @joeconwaystk ^

joeconwaystk: I think the word select sucks :)

PR #83, opened 2016-09-26
Jc/tuts

joeconwaystk:
This PR adds the tutorials back into the main branch in /doc. They have been updated for changes to the API. Additionally, a bit of renaming has gone on to make concepts/APIs clearer to understand.

This relies on #82 which relies on #81.


doc/tutorial/2016-06-23-deploying-and-other-fun-things.md
```diff
@@ -0,0 +1,200 @@
...
+```yaml
+database:
+ username: dart
+ password: dart
+ host: localhost
+ port: 5432
+ databaseName: dart_test  
+```
+
+Now, we need to do two things to make this configuration file become a reality. First, whenever we start the `Application`, we need to pass this configuration information to each pipeline. Then, the pipeline must use this information to tell the persistent store of its `ModelContext` where to connect to. (Right now, we hardcoded it to our local database for testing.)
```
smswz: still have pipeline instead of sink in text

lib/base/parameter_matching.dart
```diff
@@ -55,10 +51,14 @@ class _HTTPControllerCache {
     allDeclarations.values
         .where((decl) => decl is VariableMirror)
         .where((decl) => decl.metadata.any((im) => im.reflectee is _HTTPParameter))
-        .map((decl) => new _HTTPControllerCachedParameter(decl))
-        .forEach((_HTTPControllerCachedParameter param) {
-          if (param.httpParameter.isRequired) {
-            _hasControllerRequiredParameter = true;
+        .forEach((decl) {
+          _HTTPControllerCachedParameter param = null;
```
smswz: why initialize to null?

lib/base/parameter_matching.dart
```diff
@@ -102,35 +106,41 @@ class _HTTPControllerCache {
...
-  List<dynamic> orderedParametersFromRequest(Request req) {
-    return orderedPathParameters
-        .map((param) => _convertParameterWithMirror(req.path.variables[param.name], param.typeMirror))
-        .toList();
+  List<_HTTPControllerCachedParameter> get pathParameters => positionalParameters.where((p) => p.httpParameter is HTTPPath).toList();
+
+  List<dynamic> positionalParametersFromRequest(Request req, Map<String, List<String>> queryParameters) {
+    return positionalParameters.map((param) {
+      if (param.httpParameter is HTTPPath) {
+        return _convertParameterWithMirror(req.path.variables[param.name], param.typeMirror);
```
smswz: can this not fail?

lib/db/model.dart
```diff
@@ -77,24 +77,24 @@ class Model<PersistentType> extends Object with _QueryMatchableExtension impleme
       var property = entity.properties[k];
 
       if (property == null) {
-        throw new QueryException(400, "Key $k does not exist for ${MirrorSystem.getName(mirror.type.simpleName)}", -1);
+        throw new QueryException(QueryExceptionEvent.requestFailure, message: "Key $k does not exist for ${MirrorSystem.getName(mirror.type.simpleName)}");
```
smswz: I like this much better than magic numbers 👍

doc/tutorial/2016-06-19-executing-queries.md
```diff
@@ -0,0 +1,324 @@
...
+      "description" : "What's the tallest mountain in the world?"
+  }]));
+  expect(response.decodedBody, hasLength(1));
+});
+```
+
+This test will pass, along with the rest of them. It's important to note that GET `/questions` without a `contains` query still yields the correct results. That is because the `HTTPQuery` argument was declared in the optional parameters portion of the responder method. If the parameter were in the required, positional set of parameters and the query string was not included, this request would respond with a 400. (The same positional vs. optional behavior is true of `HTTPHeader`s as well.) For example, if we wanted to make a 'X-Client-ID' header that had to be included on this request, we'd do the following:
+
+```dart
+@httpGet getAllQuestions(@HTTPHeader("X-Client-ID") int clientID, {@HTTPQuery("contains") String contains: null}) async {
```
joeconwaystk: So there's no longer order matching for path components. There is now HTTPPath (a subclass of HTTPParameter). All positional args (header/query/path) are required. All optional args are optional.

doc/tutorial/2016-06-20-getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+author: Me
+
+environment:
+  sdk: '>=1.0.0 <2.0.0'
+
+dependencies:
+  aqueduct: any  
+```
+
+This pubspec declares an application named `quiz` (all Dart files, directories and application identifiers are snake_case), indicates that it can use a version of the Dart SDK between 1.0 and 2.0, and depends on the `aqueduct` package.
```
joeconwaystk: That's a good question... I think I may lower bound it at the current version of Dart, as there is no reason not to.

doc/tutorial/2016-06-19-executing-queries.md
```diff
@@ -0,0 +1,324 @@
...
+Run the tests again, good to go!
+
+That's fetch and insert. Delete works the same way - you specify a `predicate` or `matchOn` values and invoke `delete` on the query. If you want to update database rows, you specify both `values` and a `predicate`/`matchOn`.
+
+The more you know: Query Parameters and HTTP Headers
+---
+You can specify that a `HTTPController` responder method extract HTTP query parameters and headers and supply them as arguments to the method. We'll allow the `getAllQuestions` method to take a query parameter named `contains`. If this query parameter is passed, we'll filter the questions on whether or not that question contains a substring. In `question_controller.dart`, update this method:
+
+```dart
+@httpGet getAllQuestions({@HTTPQuery("contains") String contains: null}) async {
```
joeconwaystk: Fixed

doc/tutorial/2016-06-19-executing-queries.md
```diff
@@ -0,0 +1,324 @@
...
+    questionQuery.matchOn.description = whereContains(contains);
+  }
+  var questions = await questionQuery.fetch();
+  return new Response.ok(questions);
+}
+```
+
+Notice that the string used to match the query parameter is the first argument to `HTTPQuery`. You may name the associated variable whatever you like, it doesn't have to match the name in the query parameter. Also, note that we first check `contains` to make sure it is not-null. If we simply assigned `null` to `description`, we'd be creating a predicate that checked to see if the `description` *contained* `null`.
+
+Using HTTP header values as parameters is accomplished in the same way, except using the `HTTPHeader` metadata. Both are evaluated case-insensitively.
```
joeconwaystk: Fixed

doc/tutorial/2016-06-20-getting-started.md
joeconwaystk: Fixed

doc/tutorial/2016-06-20-getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+    "What's the tallest mountain in the world?"
+  ];
+
+  @httpGet Future<Response> getAllQuestions() async {
+    return new Response.ok(questions);
+  }
+}
+```
+
+The `QuestionController` class defines a property named `questions` that has a list of strings. Then, it defines a 'responder method' called `getAllQuestions`. Responder methods in an `HTTPController` are responsible for responding to a request. To be a responder method in an `HTTPController`, it must have `HTTPMethod` metadata that indicates which HTTP method it responds to. In this case, `getAllQuestions` will respond to `@httpGet` requests. There are built-in constants for the common HTTP methods - like `@httpPut`, `@httpPost` - which are all instances of `HTTPMethod`.
```
joeconwaystk: Calling it out as an official term, will use italics instead as that is the convention I'm going for

doc/tutorial/2016-06-20-getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+
+At the bottom of `quiz.dart`, create a `RequestSink` subclass:
+
+```dart
+class QuizSink extends RequestSink {
+  QuizSink(Map<String, dynamic> options) : super (options);
+}
+```
+
+A `RequestSink` subclass must have a constructor that takes a `Map<String, dynamic>` of options and forward it on to its superclass' constructor. These values are often provided by a configuration file and are used to configure a `RequestSink`'s properties - like database connection information. Since the `quiz` app doesn't do much right now, we simply forward the configuration options on to the `super`'s constructor as required.
```
joeconwaystk: Not according to the AP style guide :)

doc/tutorial/2016-06-20-getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+
+Both `RequestSink` and `HTTPController` are subclasses of `RequestController` - and therefore, both `QuizSink` and `QuestionController` are as well.
+A `RequestController` receives `Request`s from another `RequestController` and either responds to that request or passes it down the stream to the next `RequestController`. Thus, `RequestController`s are chained together to create a stream that a request travels in. A `RequestSink` is special in that it receives `Request`s from the `Application`, and is therefore the first controller in the stream.
+
+By default, a `RequestSink` subclass simply forwards `Request`s to its `initialController` - its `router`, which is an instance of `Router`. A `Router` splits the stream of HTTP requests based on their path. Each of these splits is called a *route*. Subsequent `RequestController`s are chained to each route. The setup for each route and the stream of `RequestController`s is defined in a `RequestSink`'s `addRoutes` required method. In this application, we'd like all HTTP requests with the path `/questions` to go to our `QuestionController`. Implement this method in `QuizSink`:
```
joeconwaystk: Fixed

doc/tutorial/2016-06-20-getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+    router
+      .route("/questions")
+      .generate(() => new QuestionController());
+  }
+}
+```
+
+There are a few things to notice here. First, a `Router` - the type of `router` - is a subclass of `RequestController` as well. This means it can receive `Request` events and either respond to them or pass them downstream. When a `Router` gets a `Request` event, it looks at all of its registered routes to find one that matches the path of the `Request`. Routes are registered through the `route` method.
+
+The `route` method method creates an instance of `RouteController` (also a `RequestController` subclass - are you seeing the pattern?). When a router matches the path of an HTTP request to one of its registered routes, the `Request` event is sent to the associated `RouteController`. If no registered route matches the path of the request, the `Router` responds to the `Request` with a 404 status code and drops the `Request` event.
```
joeconwaystk: fixed

doc/tutorial/2016-06-20-getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+
+A `RouteController` doesn't do anything with a `Request`, it simply forwards it downstream. Thus, the `addRoutes` method must chain another `RequestController` to each generated `RouteController`. In this example, the next controller is defined by the `generate` method. The `generate` method takes a closure that returns an instance of some `RequestController`. Each time the previous controller emits a `Request` event to a `generate`d stream, the closure is called, creating a new instance of a `RequestController`, and the `Request` is delivered to that new instance. Here, that new instance is an instance of our `QuestionController`.
+
+We'll get to the specifics of all of that in a moment, but we're almost to the point that we can run this web server, and that seems more exciting. With a `RequestSink` and a route hooked up to our `QuestionController`, we can write code that starts the web server. At the top of `quiz.dart`, underneath the import, define a `main` function like so:
+
+```dart
+import 'package:aqueduct/aqueduct.dart';
+
+void main() {
+  var app = new Application<QuizSink>();
```
joeconwaystk: two reasons for the change: it's a common name for Dart classes that do something similar, and there is a 'Pipeline' in the Shelf web server package that is kind of different but kind of the same

doc/tutorial/2016-06-20-getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+  var questions = [
+    "How much wood can a woodchuck chuck?",
+    "What's the tallest mountain in the world?"
+  ];
+
+  @httpGet getAllQuestions() async {
+    return new Response.ok(questions);
+  }
+
+  @httpGet getQuestionAtIndex(@HTTPPath("index") int index) async {
```
joeconwaystk: The plan is to have separate methods for each variant of the route for this controller, and so having an HTTPPath as optional should be a failure. Currently, I think it will fail to do anything (silently). I have some ideas on how to make it report an error appropriately, but it's a bigger change than for right now.

doc/tutorial/2016-06-20-writing-tests.md
```diff
@@ -0,0 +1,199 @@
...
+  var questions = [
+		"How much wood can a woodchuck chuck?",
+		"What's the tallest mountain in the world?"
+	];
+
+  @httpGet getAllQuestions() async {
+    return new Response.ok(questions);
+  }
+
+  @httpGet getQuestionAtIndex(int index) async {
```
joeconwaystk: fixed

doc/tutorial/2016-06-20-writing-tests.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+Note that we import `quiz.dart` and since the `quiz` library defined in this file exports Aqueduct, any file that imports `quiz.dart` will also import Aqueduct. Finally, get your dependencies again to get your project to recognize that `quiz` is now a library package. You can ensure that everything still works by running `bin/start.dart` again and typing a URL into your browser.
+
+Writings Tests
+---
+
+We'd like to ensure that when we hit the `/questions` endpoint, we get a response with questions. What does that mean? Well, that is up to us. But, let's say that 'questions' means 'a list of strings that all end in a question mark'.
+
+In Dart, tests are stored in a top-level `test` directory. Create that directory in `quiz`. Then, add a new file to it named `question_controller_test.dart`. (Test must end in `_test.dart` and live in the `test` directory for the tools to find them without you having to specify their path.) In this file, import both the `test` and `quiz` package.
```
joeconwaystk: fixed

doc/tutorial/2016-06-23-deploying-and-other-fun-things.md
joeconwaystk: I plan to update this chapter later once we've finished work on Heroku deploying, so for now, this chapter is marked in pivotal as needing work

lib/base/parameter_matching.dart
joeconwaystk: fixed

lib/base/parameter_matching.dart
joeconwaystk: How so?  you mean be null?

joeconwaystk:
@smswz Made changes, merge is on you if satisfactory

doc/tutorial/2016-06-20-getting-started.md
smswz: We should make a note about writing in AP style going forward 😃

lib/base/parameter_matching.dart
smswz: Yes, the HTTPQuery and HTTPHeader methods look like they can return `null`. This method can return `null`, but I'm assuming at this part in the logic it's impossible for the path parameter to be `null`. That seems like an assumption that if the code path ever changes could result in a bug here.

joeconwaystk: Ah yes, that is correct. I see what you are saying. Yeah the method is selected based on the path parameters prior to even getting into here - it's impossible to select the CachedMethod without having appropriate values for the path params. So while it's not explicitly tested or handled, all of HTTPController would come crashing down in its other tests if the case was otherwise

PR #84, opened 2016-09-27
Jc/newdriver

joeconwaystk:
Changes:

Integrated the new Postgres driver. The pubspec currently points at the git repository for now, as I'm not ready to publish the driver yet.
Implements the suggested change for decoupling QueryException from the HTTP layer.
Includes code that will prevent an isolate crash if an underlying socket gets detached or a response is sent more than once.
Some general cleanup & changes to documentation.


lib/base/request.dart
```diff
@@ -177,7 +177,7 @@ class Request implements RequestHandlerResult {
   String toDebugString({bool includeElapsedTime: true, bool includeRequestIP: true, bool includeMethod: true, bool includeResource: true, bool includeStatusCode: true, bool includeContentSize: false, bool includeHeaders: false, bool includeBody: false}) {
     var builder = new StringBuffer();
     if (includeRequestIP) {
-      builder.write("${innerRequest.connectionInfo.remoteAddress.address} ");
+      builder.write("${innerRequest.connectionInfo?.remoteAddress?.address} ");
```
rosshambrick: is the strong typing requiring this change?

joeconwaystk: No, this a bug we found in production - somehow, a particular client was disconnecting during the request. The connectionInfo was null causing a crash when we accessed its remoteAddress. There is a test that detaches a socket to ensure this behavior doesn't happen again, and this was the fix

lib/base/request_handler.dart
```diff
@@ -152,6 +147,50 @@ class RequestHandler extends Object with APIDocumentable {
     return req;
   }
 
+  void _handleError(Request request, dynamic caughtValue, StackTrace trace) {
+    try {
+      if (caughtValue is HTTPResponseException) {
+        var response = caughtValue.response;
+        _applyCORSHeadersIfNecessary(request, response);
```
rosshambrick: I feel like this should be handled at a higher level

joeconwaystk: Where is a higher level than this?

rosshambrick: I'll let you know when I learn more about how this all works 😄

rosshambrick: Essentially, it seems like there would (or could) be an opportunity to apply CORS headers, or do similar global things, in one place.

joeconwaystk: If we provide the response with information from the request handler, it could compute its own CORS headers I believe

pubspec.yaml
```diff
@@ -15,10 +15,10 @@ dependencies:
   crypto: "^2.0.0"
   logging: ">=0.11.3 <0.12.0"
   analyzer: ">=0.28.0 <0.30.0"
-  postgresql:
+  postgres:
     git:
-      url: git@github.com:stablekernel/dart_postgresql.git
-      ref: jc/modern
+      url: git@github.com:stablekernel/postgresql-dart.git
+      ref: master
```
rosshambrick: can we use the `https` version of this dependency to simplify deployments?

joeconwaystk: Well, we'll be pointing directly at pub for release, so we can defer that solution until later

lib/base/request_handler.dart
rosshambrick: what about some post-processing type middleware?

rosshambrick:
👍

joeconwaystk:
@rosshambrick Good to go from your perspective? Merge is on you

smswz:
Got merge conflicts now :/

joeconwaystk:
should just be renamed, will check them out now

joeconwaystk:
Welp, I have no idea how that removed all changed files.

rosshambrick:
Did 1.0 already have these changes somehow?

joeconwaystk:
Maybe? I know both Jesse and I have done some rebasing/merging on various branches, but this still surprises me

lib/base/request_handler.dart
rosshambrick: I was thinking more of a global hook for all requests

rosshambrick:
so if nothing changed, it's good to merge right?

joeconwaystk:
yeah, go for it 👍

PR #85, opened 2016-09-29
Jc/robust

joeconwaystk:
This PR includes a number of robust tests for executing queries, especially around join queries. And addresses a bug or two along the way.

### DIFF
--- a/doc/tutorial/2016-06-19-executing-queries.md
+++ b/doc/tutorial/2016-06-19-executing-queries.md
@@ -59,11 +59,11 @@ Why there is a persistent and instance type we'll dig into later, but it is impo
 Defining a Context
 ---
 
-In order for an application to work with a database, it needs a `ModelContext`. A `ModelContext` is the facilitator between your code, model objects and a database. It is made up of two components, a `DataModel` (the thing that keeps track of all of your model type) and `PersistentStore` (the thing that talks to the database). These objects are set up in a `RequestSink`. In `quiz_sink.dart`, add the following code to the constructor for `QuizSink` and define a new property:
+In order for an application to work with a database, it needs a `ModelContext`. A `ModelContext` is the facilitator between your code, model objects and a database. It is made up of two components, a `DataModel` (the thing that keeps track of all of your model types) and `PersistentStore` (the thing that talks to the database). These objects are set up in a `RequestSink`. In `quiz_request_sink.dart`, add the following code to the constructor for `QuizRequestSink` and define a new property:
 
 ```dart
-class QuizSink extends RequestSink {
-  QuizSink(Map<String, dynamic> options) : super(options) {
+class QuizRequestSink extends RequestSink {
+  QuizRequestSink(Map<String, dynamic> options) : super(options) {
     var dataModel = new DataModel([Question]);
     var persistentStore = new PostgreSQLPersistentStore.fromConnectionInfo("dart", "dart", "localhost", 5432, "dart_test");
     context = new ModelContext(dataModel, persistentStore);
@@ -136,7 +136,7 @@ setUpAll(() async {
 });
 ```
 
-After the application is started, we know that it creates a `ModelContext` in the constructor of `QuizSink`. The default context can be accessed through `ModelContext.defaultContext`. We also know that this context has a `DataModel` containing `Question`. The class `SchemaGenerator` will create a database-agnostic JSON schema file from the `DataModel`. Subclasses of `SchemaGeneratorBackend`, like `PostgreSQLSchemaGenerator`, can take that JSON and create SQL commands that create all of the tables, indices and constraints defined by the JSON schema. Each of those is executed on the context, creating the schema in the database. (Note that the `temporary` flag adds makes all of the tables temporary and therefore they disappear when the database connection in the context's persistent store closes.)
+After the application is started, we know that it creates a `ModelContext` in the constructor of `QuizRequestSink`. The default context can be accessed through `ModelContext.defaultContext`. We also know that this context has a `DataModel` containing `Question`. The class `SchemaGenerator` will create a database-agnostic JSON schema file from the `DataModel`. Subclasses of `SchemaGeneratorBackend`, like `PostgreSQLSchemaGenerator`, can take that JSON and create SQL commands that create all of the tables, indices and constraints defined by the JSON schema. Each of those is executed on the context, creating the schema in the database. (Note that the `temporary` flag adds makes all of the tables temporary and therefore they disappear when the database connection in the context's persistent store closes.)
 
 Because the `PostgreSQLPersistentStore`'s connection to the database is also a stream, it, too, must be closed to let the test's main function terminate. In `tearDownAll`, add this code before the app is terminated:
 
--- a/test/db/postgresql/fetch_test.dart
+++ b/test/db/postgresql/fetch_test.dart
@@ -13,10 +13,7 @@ void main() {
   test("Fetching an object gets entire object", () async {
     context = await contextWithModels([TestModel]);
 
-    var m = new TestModel()
-      ..name = "Joe"
-      ..email = "a@a.com";
-
+    var m = new TestModel(name: "Joe", email: "a@a.com");
     var req = new Query<TestModel>()..values = m;
     var item = await req.insert();
 
@@ -31,9 +28,7 @@ void main() {
   test("Specifying resultProperties works", () async {
     context = await contextWithModels([TestModel]);
 
-    var m = new TestModel()
-      ..name = "Joe"
-      ..email = "b@a.com";
+    var m = new TestModel(name: "Joe", email: "b@a.com");
     var req = new Query<TestModel>()..values = m;
 
     var item = await req.insert();
@@ -54,9 +49,7 @@ void main() {
     context = await contextWithModels([TestModel]);
 
     for (int i = 0; i < 10; i++) {
-      var m = new TestModel()
-        ..name = "Joe${i}"
-        ..email = "asc${i}@a.com";
+      var m = new TestModel(name: "Joe${i}", email: "asc${i}@a.com");
       var req = new Query<TestModel>()..values = m;
       await req.insert();
     }
@@ -91,9 +84,7 @@ void main() {
     context = await contextWithModels([TestModel]);
 
     for (int i = 0; i < 10; i++) {
-      var m = new TestModel()
-        ..name = "Joe${i}"
-        ..email = "desc${i}@a.com";
+      var m = new TestModel(name: "Joe${i}", email: "desc${i}@a.com");
 
       var req = new Query<TestModel>()..values = m;
 
@@ -117,9 +108,7 @@ void main() {
     context = await contextWithModels([TestModel]);
 
     for (int i = 0; i < 10; i++) {
-      var m = new TestModel()
-        ..name = "Joe${i%2}"
-        ..email = "multi${i}@a.com";
+      var m = new TestModel(name: "Joe${i%2}", email: "multi${i}@a.com");
 
       var req = new Query<TestModel>()..values = m;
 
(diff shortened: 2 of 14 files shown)


joeconwaystk:
I'll rebase this one and it'll be much shorter after other PRs are merged

doc/tutorial/2016-06-19-executing-queries.md
```diff
@@ -0,0 +1,325 @@
...
+
+*All models must have a primary key.* Other interesting flags are `indexed`, `nullable` and `defaultValue`. If a property does not have a `ColumnAttributes`, it is still a persistent property, it's just a normal column and its database type is derived from its Dart type. Supported Dart types are `int`, `double`, `String`, `DateTime` and `bool`.
+
+Once a persistent type has been defined, you must define an instance type. The instance type is a subclass of `Model` and must implement its associated persistent type and also provide it as a type argument to `Model<T>`. At the top of this file, but underneath the part of directive, add the following:
+
+```dart
+part of quiz;
+
+class Question extends Model<_Question> implements _Question {}
+class _Question {
```
rosshambrick: Remind me: Is this double class thing to allow partial queries?  And if we required all or nothing it could go away?

doc/tutorial/2016-06-19-executing-queries.md
```diff
@@ -0,0 +1,325 @@
...
+  String description;
+}
+```
+
+Why there is a persistent and instance type we'll dig into later, but it is important. We now have a model named `Question`. We need to set up the application to interact with a database.
+
+Defining a Context
+---
+
+In order for an application to work with a database, it needs a `ModelContext`. A `ModelContext` is the facilitator between your code, model objects and a database. It is made up of two components, a `DataModel` (the thing that keeps track of all of your model type) and `PersistentStore` (the thing that talks to the database). These objects are set up in a `RequestSink`. In `quiz_sink.dart`, add the following code to the constructor for `QuizSink` and define a new property:
```
rosshambrick: `all of your model type` -> `all of your model types`

doc/tutorial/2016-06-19-executing-queries.md
joeconwaystk: It serves a few purposes now. It allows for transient property distinction by having the two different classes. Saves a lot of keystrokes because you don't have to mark every DB-backed property with attributes. It does allow for a sparse map. It also is the driver for building matcher queries and providing values for queries.

doc/tutorial/2016-06-19-executing-queries.md
```diff
@@ -0,0 +1,325 @@
...
+
+```dart
+class QuizSink extends RequestSink {
+  QuizSink(Map<String, dynamic> options) : super(options) {
+    var dataModel = new DataModel([Question]);
+    var persistentStore = new PostgreSQLPersistentStore.fromConnectionInfo("dart", "dart", "localhost", 5432, "dart_test");
+    context = new ModelContext(dataModel, persistentStore);
+  }
+
+  ModelContext context;
```
rosshambrick: the field declaration should go above the constructor right?

joeconwaystk: I've been doing it the other way. I'm still trying to settle on an appropriate order for class files, up for suggestions. Currently, I go static methods first, constructors, properties, public methods, private methods.

doc/tutorial/2016-06-19-executing-queries.md
```diff
@@ -0,0 +1,325 @@
...
+
+Why there is a persistent and instance type we'll dig into later, but it is important. We now have a model named `Question`. We need to set up the application to interact with a database.
+
+Defining a Context
+---
+
+In order for an application to work with a database, it needs a `ModelContext`. A `ModelContext` is the facilitator between your code, model objects and a database. It is made up of two components, a `DataModel` (the thing that keeps track of all of your model type) and `PersistentStore` (the thing that talks to the database). These objects are set up in a `RequestSink`. In `quiz_sink.dart`, add the following code to the constructor for `QuizSink` and define a new property:
+
+```dart
+class QuizSink extends RequestSink {
```
rosshambrick: `QuizSink` -> `QuizRequestSink`

doc/tutorial/2016-06-19-executing-queries.md
```diff
@@ -0,0 +1,325 @@
...
+  ...
+```
+
+(In the future, we'll allow this information to be passed from a configuration file. But for now, we'll do it manually.)
+
+A `DataModel` is initialized with a list of all instance types in your application. The persistent store is a specific implementation of a persistent store, `PostgreSQLPersistentStore`. It is initialized with information necessary to connect to a database. A `ModelContext` simply contains those two things to tie them together.
+
+(By the way, the interface for `PersistentStore` can be implemented for different flavors of SQL and even non-SQL databases. We just so happen to prefer PostgreSQL, so we've already built that one.)
+
+When a `ModelContext` is created, it becomes the *default context* of your application. When we execute database queries, they default to this default context. If we have multiple databases, we can create more `ModelContext`s and pass them around to make sure we hit the right database. For now, we can ignore this, just know that it exists.
```
rosshambrick: Would this be a good use case for a read-only context that you might use for reporting?

joeconwaystk: Yeah, the context's persistent store is bound to an authenticated user to the DB that has privileges, so yes, i like that idea.

joeconwaystk: It's really under tested right now tho so I'm not pressing it hard

doc/tutorial/2016-06-19-executing-queries.md
```diff
@@ -0,0 +1,325 @@
...
+Executing Queries
+---
+
+Now that we have a context - which can establish a connection to a database, talk to the database and map rows to model objects - we can execute queries in our `RequestController`s. In `question_controller.dart`, remove the list of static questions and replace the code for `getAllQuestions` (we'll work on `getQuestionAtIndex` soon):
+
+```dart
+class QuestionController extends HttpController {
+  @httpGet getAllQuestions() async {
+    var questionQuery = new Query<Question>();
+    var questions = await questionQuery.fetch();
```
rosshambrick: maybe `fetchAll()` would be a better name for this method?

joeconwaystk: It gets tough because I might look at 'fetchAll' and expect to get all objects in an entity/table, but there may be a predicate applied.

joeconwaystk: And then there is a consistency to the naming... does 'deleteAll' or 'updateAll' sound better?

rosshambrick: ah, the filtering is added to the query right?  not the fetch

joeconwaystk: Right - query is populated with values, and then gets execute thru some operation. So technically you could reuse a query for multiple operations. Like a fetch, then a delete

rosshambrick: what if instead of a Query fetching all by default, it's more of a builder pattern and you have to do something like this to make it explecit:
`var allQuestions = new Query<Question>().all();

joeconwaystk: What's the default do then? Also, the builder pattern is ugly as shit, and cascading operators are preferred here

rosshambrick: The default wouldn't do anything, it would be a builder.  Something like:
```
Query<Question> allQuestions = new QueryBuilder<Question>().all();

Query<Question> mcQuestions = new QueryBuilder<Question>()
    .where(q => q.isMultipleChoice)
    .build();
```
I'm just trying to avoid a default that could cause problems unintentionally.  Like if it defaults to getting all, that could be really slow or have memory implications if not explicit.

No worries for not, just thinking through things.

test/db/postgresql/fetch_test.dart
```diff
@@ -260,6 +258,80 @@ void main() {
     expect(fResult.backingMap["text"], isNull);
   });
 
+  test("Fetch one that returns more than one row (because of join) throws exception", () async {
+    context = await contextWithModels([GenUser, GenPost]);
+
+    var objects = [
+      new GenUser()
+        ..name = "Joe",
```
rosshambrick: I feel like named optional parameters would read better than this.  Can't you do something like:
```
new GenUser(name: "Joe");
```

joeconwaystk: yeah, if you declared a constructor like so:

```
GenUser({this.name});
```

test/db/postgresql/has_many_fetch_test.dart
```diff
@@ -3,491 +3,568 @@ import 'package:aqueduct/aqueduct.dart';
...
-        return pairs.map((pair) {
-          var q = new Query<Equipment>()
-            ..values.name = pair.first
-            ..values.type = pair.last
-            ..values.location = (new Location()..id = loc.id);
-          return q.insert();
-        });
+    test("Fetch has-many relationship that is empty returns empty, and deeper nested relationships are ignored even when included", () async {
+      var q = new Query<Parent>()
+        ..matchOn.children.includeInResultSet = true
```
joeconwaystk: In this case, includes `children` in the result set. Aka, joins the Parent table with Child table

doc/tutorial/2016-06-19-executing-queries.md
joeconwaystk: Well, if someone leaves a Query in untested that fetches an entire table, I think that's on them :)

joeconwaystk: And to give you the versions of those two queries:

```
var allQuestions = new Query<Question>();

var mcQuestions = new Query<Question>()
  ..matchOn.isMultipleChoice = true;
```

test/db/postgresql/has_many_fetch_test.dart
rosshambrick: hmm...so the `matchOn` confuses me then.  Could it be something like this?
```dart
var q = new Query<Parent>()
    ..eagerLoad.children;
```

rosshambrick: or:
```dart
var q = new Query<Parent>()
    ..eagerLoad.children.eagerLoad.toy.and.vaccinations
    ..where.name = "D";
```

joeconwaystk: Yeah, I tried something like that. It doesn't work well. That first line in particular doesn't really indicate anything as there are no values being set, just sort of calling getters. While it reads like a sentence, it doesn't make sense in the scope of the programming language. Why would a getter mutate the state of the query? And then you get into nested relationships as well. I opted to be very explicit after a few iterations.

joeconwaystk: w.r.t to 'where' vs 'matchOn'. The matchers themselves all begin with 'where' as a way to namespace them. Also, I had the split before like you have now between 'where' and 'eagerLoad'. It was 'matchOn' and 'include'. I didn't like it, and it really made the query building code way more complex, both internally and in using it.

rosshambrick:
There are 3 tests that fail for me.  They are all similar to this:
```
dart:isolate  _RawReceivePortImpl._handleMessage

Failed to load "/Users/ross/src/aqueduct/example/templates/default/test/register_test.dart":
Unable to spawn isolate: Unhandled exception:
Could not import "package:wildfire/wildfire.dart" from "file:///Users/ross/src/aqueduct/example/templates/default/test/register_test.dart": No mapping for 'wildfire' package when resolving 'package:wildfire/wildfire.dart'.
null
```

joeconwaystk:
Weird, where are you running tests from/command you are using?

joeconwaystk:
Those are the templates tests that aren't really meant to run on their own - they get added with the template. Just not sure how they got ran

doc/tutorial/2016-06-19-executing-queries.md
joeconwaystk: fixed

doc/tutorial/2016-06-19-executing-queries.md
joeconwaystk: fixed

test/db/postgresql/fetch_test.dart
joeconwaystk: I started doing this for a few, but man it is painstaking - will defer.

PR #90, opened 2016-10-03
Adds refresh_token as grant_type to conform to RFC 6749

evils:
**Changes:**
* Accepts refresh_token as grant_type as described in RFC 6749 which describes OAuth2

* Tests included


anachlas:
👍

PR #91, opened 2016-10-14
Jc/migration

joeconwaystk:
Database migration. Didn't quite get all the way there - big missing thing is generating migration files that have a diff of the current schema to previous schema, but the foundation is there.

PR #92, opened 2016-10-17
API reference documentation and some name changes

joeconwaystk:
Sorry for the size. Changed a lot of names to make the API more discoverable. Adding documentation for aqueduct classes and methods. This branch needs to be rebased after #91 is merged.


smswz:
Are there any non-name changes? Github chokes on this diff

joeconwaystk:
Well, name changes and API docs. If you checkout this branch and run dartdoc, then open doc/api/index.html, you'll see the full breadth of the changes

smswz:
👍  from me

anachlas:
The only one I'm not sure about is `OrderedSet` -> `ManagedSet`.  Thoughts?

joeconwaystk:
Really? Why? That was the one I was most sure of. Not saying I'm right, but interested as to why

anachlas:
Well set implies to me an unordered collection, but it looks like it would preserve order.

anachlas:
Also set implies uniqueness which isn't necessary here.

anachlas:
Well actually it is unique, but the ordered thing still bugs me.

joeconwaystk:
Well it is unique, but also there is no order - it'll be in the order the database decides. It may even be a subset.

anachlas:
Ok in that case I'm fine with it, 👍

PR #102, opened 2016-10-31
Fix Type Checking for Transient Properties

MarcusSmith:


test/db/data_model_test.dart
```diff
@@ -379,4 +388,12 @@ class _InvalidModel {
   @managedPrimaryKey int id;
 
   Uri uri;
```
joeconwaystk: Sweet, looks like we were covering persistent properties but not transient.

test/db/model_test.dart
```diff
@@ -368,6 +368,35 @@ void main() {
     expect(m["bothOverQualified"], "foo");
   });
 
+  test("Transient Properties of all types can be read and returned", () {
+
+    var dateString = "2016-10-31T15:40:45+00:00";
+
+    var m = (new TransientTypeTest()..readMap({
```
joeconwaystk: Can we shorten this test so that we declare the map of values first, drop it into readMap, and then do an expect(inputMap, m.asMap())?

PR #103, opened 2016-11-01

ronlobo:
Git remove migration files.


joeconwaystk:
I agree that migration.yaml should be ignored - this has sensitive info - but the *.migration.dart files should be checked into source control. Can you update that?

ronlobo:
Sure

ronlobo:
Reopening the PR

PR #104, opened 2016-11-01

ronlobo:
Git remove migration.yaml.
Git remove error.log.
Add migration default file.
Extend docs.


ronlobo:
Adding more stuff to gitignore

PR #105, opened 2016-11-01

ronlobo:
Git remove error.log.
Git remove migration.yaml. Add migration default file. Extend docs.
Git remove .aqueductpid, .aqueductsignal and all api.log.*.
Add description for default config.


ronlobo:
one more...

Issue #101, opened 2016-10-30
Error when creating a new project from command line on Windows 7


aelayeb:
When I'm trying to create a new project from command line with the "aqueduct create" command, I'm getting an error on Windows 7.

Here is the output :

```
C:\Users\Aelayeb>aqueduct create -n web_dart
Fetching Aqueduct as:
  aqueduct: "any"
Determining Aqueduct template source...
ProcessException: Le fichier sp├®cifi├® est introuvable.

  Command: pub get --no-packages-dir
```

Tested on Windows 7 64bits.

joeconwaystk:
So forgive me for not understanding a thing about Windows, but this would seem to indicate that `pub` is not in your PATH. What happens when you run `pub get` from the command line by itself?

aelayeb:
pub is in my PATH and it works fine, but it seems that the "aqueduct create" fails to switch directory before making the pub get command.

joeconwaystk:
So, it must be here: https://github.com/stablekernel/aqueduct/blob/master/lib/commands/template_creator.dart#L75

The template generator works by creating a directory and putting a pubspec.yaml file in that directory. Then it runs pub get to fetch the entire package, which contains the templates. Then it effectively copies over the template contents to that directory. It's either having problems with setting the CWD to the project directory, or finding the pubspec.yaml file in the project directory.

Would you be able to step through the template generator code on your machine? If you instead clone this repository and reactive aqueduct to point at the local installation (`pub global activate -s path path/to/aqueduct`), you should be able to dig into the state of the filesystem at this point.

aelayeb:
Yeah it seems the problem come from this line : 
`var result = Process.runSync("pub", ["get", "--no-packages-dir"], workingDirectory: projectDirectory.path);`
in template_creator.dart.

I tried to do somehing with a minimal code and it failed with the same error.
```
import 'dart:io';

main () {
	var result = Process.runSync("dir", []);
    if (result.exitCode != 0) {
		throw new Exception("${result.stderr}");
    }
}
```
("dir" command is equivalent to "ls" on linux)
It's as if the "Process.runSync" method doesn't use the PATH on windows.
I think it's a dart problem, not an aqueduct one.

aelayeb:
When I add the "runInShell: true" parameter its seems to work but I dont know why it works like this on windows...

joeconwaystk:
[...]

PR #111, opened 2016-11-03
Fix path error when creating new project on Windows, see #101

aelayeb:


aelayeb:
Maybe we need to add "path" package to dependencies in the pubspec ?

joeconwaystk:
This works! 

re: Path - Yes, even though yaml depends on it, I think if we explicitly refer to functionality in path it should be in the pubspec. Looks like "^1.0.0" will work.

joeconwaystk:
Once path is added to pubspec, I'll merge this into my branch and we'll get this deployed to pub shortly, with a few other fixes that have gone on.

aelayeb:
Ok good !
One more thing : I was careless about naming the path import.
It'll be better if you can rename the "pathLib' occurences to "path_lib".
According to : https://www.dartlang.org/guides/language/effective-dart/style#do-name-import-prefixes-using-lowercasewithunderscores

PR #118, opened 2016-11-06
allow 'aqueduct setup' to work in Windows and pay attention to -u

RobertMcDermot:
The `which` command is equivalent to the `where` command in Windows. Also made some adjustments to the way the path is constructed to make it more platform independent. Further, the -u parameter was not being forwarded to the psql commands as the user might expect.

### DIFF
--- a/bin/aqueduct.dart
+++ b/bin/aqueduct.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'package:args/args.dart';
 import 'package:aqueduct/aqueduct.dart';
 
+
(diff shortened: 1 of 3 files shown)


joeconwaystk:
Awesome work. Will get a chance to review in a bit and test on macOS

JesseBlack82:
@joeconwaystk I verified that:

- `aqueduct create` still work on macs 
- `aqueduct setup` still work on macs
- Aqueduct tests are all passing (431 passed)

PR #123, opened 2016-11-09
Update changelog

JesseBlack82:
Update changelog to reflect changes from #118


anachlas:
I feel like we don't need a version bump for this, but if you do want to change the version you have to modify the pubspec as well.

JesseBlack82:
It's in preparation for the patch bump that is coming when Joe integrates some pending changes he has.  We aren't going to release this change until his work is merged as well.

PR #124, opened 2016-11-10
Jc/cors

joeconwaystk:
This PR:

1. Updates CORS tests and improves behavior according to the specification.
2. Changes behavior of unimplemented HTTP methods in HTTPController to return a 405 status code with Allow header instead of a 404.

### DIFF
--- a/lib/http/router.dart
+++ b/lib/http/router.dart
@@ -19,6 +19,7 @@ class Router extends RequestController {
   /// Creates a new [Router].
   Router() {
     unhandledRequestController = _handleUnhandledRequest;
+    policy.allowCredentials = false;
   }
 
   List<RouteController> _routeControllers = [];
--- a/test/base/cors_test.dart
+++ b/test/base/cors_test.dart
@@ -24,25 +24,25 @@ void main() {
     test("Controller with no policy returns correctly", () async {
       var resp = await http.get("http://localhost:8000/nopolicy");
       expect(resp.statusCode, 200);
-      expectThatNoCORSProcessingOccured(resp);
+      expectThatNoCORSProcessingOccurred(resp);
     });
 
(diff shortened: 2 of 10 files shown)


test/base/cors_test.dart
```diff
@@ -5,202 +5,477 @@ import 'package:http/http.dart' as http;
...
 
-      // Should return 401, the preflight is unsigned and therefore doesn't make it past the authenticator, since it
-      // won't allow OPTIONS thru because it isn't expecting a preflight request.
-      expect(resp.statusCode, 401);
+      expect(resp.statusCode, 200);
+      expect(resp.headers["access-control-allow-origin"], isNull);
+      expect(resp.headers["access-control-allow-headers"], isNull);
+      expect(resp.headers["access-control-allow-methods"], isNull);
+      expect(resp.headers["access-control-expose-headers"], isNull);
+      expect(resp.headers["access-control-allow-credentials"], isNull);
```
anachlas: Can use the convenience `expectThatNoCORSProcessingOccured` here.

joeconwaystk: Technically, I can't because these are an OPTIONS and it is a GET, but I can just make an OPTIONS version

joeconwaystk: gah, nvm, I forgot what that method did.

lib/http/router.dart
```diff
@@ -19,6 +19,7 @@ class Router extends RequestController {
   /// Creates a new [Router].
   Router() {
     unhandledRequestController = _handleUnhandledRequest;
+    policy.allowCredentials = false;
```
JesseBlack82: Why do you override the CORS default here?

JesseBlack82:
👍

Issue #129, opened 2016-11-16
Response.notFound() and 404, logging


jkleiser:
In my generated test project I have commented out the `.pipe(new Authorizer(authenticationServer))` on the line under `.route("/users/[:id]")` (in my lib/src/aque_test_sink.dart) so I won't get the {"error":"No authorization header."} when I try `localhost:8000/users`. Now I just get an empty page instead. I would have expected a 404. Do I have to provide something for a Response.notFound() to give me a visible 404 page?
What's the easiest way to insert logging lines into a file like lib/src/controller/user_controller.dart?

joeconwaystk:
Aqueduct doesn't provide any behavior for HTML templating. There are other tools that can be used with Aqueduct to provide that behavior.

If you want an endpoint to return HTML, you can configure that by setting the content type of the response (or in the case of an HTTPController, setting the responseContentType). For a router, you can set the `unhandledRequestController` property to a closure that responds to the request with a 404 and html body:

```
router.unhandledRequestController = (req) {
  req.respond(new Response.notFound(headers: {"Content-Type" : "application/html"}, body: htmlString));
};
```

Any `RequestController` subclass has a `logger` property that can be used to log:

```
logger.info("A log statement in a controller");
```

jkleiser:
I didn't ask for HTML templating. I just wanted to produce a very simple HTML response. This far I have not succeeded in doing that. A tutorial example on how to do this would help me. It would also be nice to deliver static HTML files. If other tools are needed for this kind of behavior, it would be nice if the docs had some suggestions.

jkleiser:
In the router.unhandledRequestController above, I guess it should have been `{"Content-Type" : "text/html"}`.

PR #130, opened 2016-11-19
Jc/content type

joeconwaystk:
1. Adds contentType property to Request. Makes setting the content type a lot more obvious than just including it as a header.
2. Allows a Response to override the responseContentType of an HTTPController if explicitly set.
3. Allows default content type of all Responses to be set by the user.
4. Made text/* a default encoder and allows for match-all character to be used in a content-type encoder key.

This should address issues uncovered by #129

### DIFF
--- a/lib/http/http_controller.dart
+++ b/lib/http/http_controller.dart
@@ -46,11 +46,10 @@ abstract class HTTPController extends RequestController {
     _applicationWWWFormURLEncodedContentType
   ];
 
-  /// The content type of responses from this [HTTPController].
+  /// The default content type of responses from this [HTTPController].
   ///
-  /// This type will automatically be written to this response's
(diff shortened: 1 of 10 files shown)


lib/http/http_controller.dart
```diff
@@ -131,7 +131,10 @@ abstract class HTTPController extends RequestController {
     ).reflectee as Future<Response>;
```
JesseBlack82: ^^^ The documentation for `responseContentType` needs updating:

> /// The content type of responses from this [HTTPController].
   ///
   /// This type will automatically be written to this response's
   /// HTTP header. Defaults to "application/json". This value determines how the body data returned from this controller
   /// in a [Response] is encoded.
   ContentType responseContentType = ContentType.JSON;

to reflect that the responseContentType is only used when a response doesn't have an content type explicitly set

PR #136, opened 2016-11-22
Updated readme

erikist:
Added a warning about using the `./wildfire start` command which has side effects such as switching your branch to master and moving to the HEAD of master and added explicit local instructions

### DIFF
--- a/example/templates/default/README.md
+++ b/example/templates/default/README.md
@@ -61,14 +61,6 @@ aqueduct db validate
 
 ## Running wildfire
 
-#### ...locally
-
-```
-  dart bin/start.dart 
-```
-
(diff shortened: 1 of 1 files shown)


example/templates/default/README.md
```diff
@@ -61,6 +61,14 @@ aqueduct db validate
 
 ## Running wildfire
 
+#### ...locally
```
joeconwaystk: config file gets used locally too, so make sure that is mentioned, and make sure you run from the project directory

erikist:
@joeconwaystk updated it with some reasoning given.

smswz:
@joeconwaystk @erikist can this be merged?

PR #170, opened 2017-01-06
Handle the doubles that are nums

erikist:


test/db/model_test.dart
```diff
@@ -176,6 +176,15 @@ void main() {
     expect(remap["dateCreated"], dateString);
   });
 
+  test("Handles input of type num for double precision float properties of the model", () {
+    var m = (new TransientTypeTest()
```
anachlas: The outer parentheses here are unnecessary

anachlas:
👍

PR #171, opened 2017-01-09
Lowercase headers #136218279

joeconwaystk:
This PR lowercases headers when set on a Response object. Makes it so things inspecting those headers won't get confused by case-sensitivity.

### DIFF
--- a/lib/src/http/request.dart
+++ b/lib/src/http/request.dart
@@ -152,6 +152,20 @@ class Request implements RequestControllerEvent {
     }
 
     if (encodedBody != null) {
+//      ContentType type;
+//
+//      if (!responseObject.hasExplicitlySetContentType) {
+//        var contentTypeHeader = responseObject.headers[HttpHeaders.CONTENT_TYPE];
+//
+//        if (contentTypeHeader is ContentType) {
+//          type = contentTypeHeader;
+//        } else if (contentTypeHeader is String) {
+//          type = ContentType.parse(contentTypeHeader);
+//        }
+//      } else {
+//        type = responseObject.contentType;
+//      }
+
       response.headers
           .add(HttpHeaders.CONTENT_TYPE, responseObject.contentType.toString());
       response.write(encodedBody);
--- a/lib/src/http/request.dart
+++ b/lib/src/http/request.dart
@@ -152,20 +152,6 @@ class Request implements RequestControllerEvent {
     }
 
     if (encodedBody != null) {
-//      ContentType type;
-//
-//      if (!responseObject.hasExplicitlySetContentType) {
-//        var contentTypeHeader = responseObject.headers[HttpHeaders.CONTENT_TYPE];
-//
-//        if (contentTypeHeader is ContentType) {
-//          type = contentTypeHeader;
-//        } else if (contentTypeHeader is String) {
-//          type = ContentType.parse(contentTypeHeader);
-//        }
-//      } else {
-//        type = responseObject.contentType;
-//      }
-
       response.headers
           .add(HttpHeaders.CONTENT_TYPE, responseObject.contentType.toString());
       response.write(encodedBody);
--- a/lib/src/http/response.dart
+++ b/lib/src/http/response.dart
@@ -101,13 +101,12 @@ class Response implements RequestControllerEvent {
 
   /// Map of headers to send in this response.
   ///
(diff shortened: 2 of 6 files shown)


test/base/response_test.dart
```diff
@@ -0,0 +1,51 @@
+import 'package:test/test.dart';
+import 'package:aqueduct/aqueduct.dart';
+
+void main() {
+  test("Modifying return alue from Response.headers does not impact actual headers", () {
```
anachlas: Typo `value`

lib/src/http/response.dart
```diff
@@ -107,7 +107,14 @@ class Response implements RequestControllerEvent {
...
   ///
   /// Adding a Content-Type header through this property has no effect. Use [contentType] instead.
-  Map<String, dynamic> headers;
+  void set headers(Map<String, dynamic> h) {
+    _headers = {};
+    h.forEach((k, v) {
+      _headers[k.toLowerCase()] = v;
+    });
+  }
+  Map<String, dynamic> get headers => new Map.from(_headers);
```
anachlas: Maybe we could change this to something like 
`Map<String, dynamic> get headers => new Map.unmodifiable(_headers);`?

joeconwaystk: Hmm yeah. Let me take a bit to think about it - let me know if you have any ideas. The response.headers["foo"] = "bar" think stands out. I may create a private 'LowercasingKeyMap' or something and have the type of headers be that, but the surface it appears as just a normal Map

smswz: can you not use the `addHeader` method for that case?

lib/src/http/response.dart
```diff
@@ -214,6 +217,13 @@ class Response implements RequestControllerEvent {
   Response.serverError({Map<String, dynamic> headers, dynamic body})
       : this(HttpStatus.INTERNAL_SERVER_ERROR, headers, body);
 
+  void addHeader(String header, dynamic value) {
+    if (_headers == null) {
```
smswz: can _headers ever be null?

lib/src/http/response.dart
joeconwaystk: I think the issue is that it is confusing - someone would probably expect a map to work like a map and not have to go and find another method. There wouldn't be any warning that it didn't work either

lib/src/http/response.dart
joeconwaystk: Yeah if specifically null is passed in the constructor I think

anachlas: No even if you pass `null` to one of the constructors it calls the default constructor which sets it to `{}`, which is important because if you ever accessed `headers` and `_headers` was null you would get an error.

lib/src/utilities/lowercasing_map.dart
```diff
@@ -0,0 +1,32 @@
+import 'dart:collection';
+
+class LowercaseMap<V> extends Object with MapMixin<String, V> {
```
anachlas: `MapMixin` is really cool!

lib/src/http/response.dart
```diff
@@ -141,12 +139,8 @@ class Response implements RequestControllerEvent {
   /// and you should prefer to use those. Always use lowercase keys for headers.
   Response(int statusCode, Map<String, dynamic> headers, dynamic body) {
     this.body = body;
-    this.headers = headers;
+    this.headers = new LowercaseMap.fromMap(headers);
```
anachlas: I think this will crash if we pass `null` for the headers. Since its the constructor should we set the `_headers` instead?

joeconwaystk: yup, it sure did when i added another test. its now headers ?? {}

lib/src/http/request.dart
```diff
@@ -152,6 +152,20 @@ class Request implements RequestControllerEvent {
...
+//        var contentTypeHeader = responseObject.headers[HttpHeaders.CONTENT_TYPE];
+//
+//        if (contentTypeHeader is ContentType) {
+//          type = contentTypeHeader;
+//        } else if (contentTypeHeader is String) {
+//          type = ContentType.parse(contentTypeHeader);
+//        }
+//      } else {
+//        type = responseObject.contentType;
+//      }
```
anachlas: Why is this all commented out?

joeconwaystk: whoops I'm being sloppy

PR #172, opened 2017-01-09
Fixed bug where auth code flow duplicated state

joeconwaystk:
State param would being included twice in redirect url


lib/src/auth/auth_code_controller.dart
```diff
@@ -191,6 +190,7 @@ class AuthCodeController extends HTTPController {
     var redirectURI = Uri.parse(uriString);
     Map<String, String> queryParameters =
         new Map.from(redirectURI.queryParameters);
+    print("${queryParameters}");
```
anachlas: Print statement.

anachlas:
👍

PR #178, opened 2017-01-25
Renaming 'matchOn' for Query to 'where'

joeconwaystk:
First of a number of name changes


anachlas:
The only name change I would bundle in with this is maybe rename `whereRelatedByValue` to `relatedByValue` or something similar so you dont have `where.x = whereRelatedByValue(y)`

joeconwaystk:
Will be renaming matchOn in ManagedSet and will be renaming where* to something else... have a few more things to change to better figure out what to name it

PR #199, opened 2017-02-13
Add generics to where methods

joeconwaystk:
DO NOT MERGE.

Must be rebased against #198 if merged.

Adds generics to all where* methods (e.g., whereEqualTo). This gives the analyzer the ability to check types on the arguments for where methods against the property type. 

~It may be a bug in Dart generics that this is allowed in checked mode.~ Current generic implementation treats type arguments as dynamic (see https://github.com/dart-lang/sdk/issues/27460). Adding generics and casting the return type to dynamic shuts the analyzer up and at least confuses checked mode at runtime. May be a bug because `whereContains`, `whereStartsWith` and `whereEndsWith` - which must take/return a `String` do fail at runtime if declared statically. For example, the following fails at runtime in checked mode:

```
String whereContains(String string) {
  NotAString x = ...;
  return x as dynamic;
}
```

But if declared as follows, which is effectively the same thing, it passes at runtime:

```
T whereContains<T extends String>(T string) {
  NotAString x = ...;
  return x as dynamic;
}
```


joeconwaystk:
This may not work in future. Closing for now.

PR #202, opened 2017-02-14
Change propertiesToFetch to mirror sortBy, pageBy, etc.

joeconwaystk:
Rebase w/ #201.

```
var q = new Query<User>()
  ..returningProperties((t) => [t.id, t.name, t.email]);
```

Removes referencing properties by their string name.

### DIFF
--- a/test/db/postgresql/fetch_test.dart
+++ b/test/db/postgresql/fetch_test.dart
@@ -35,7 +35,7 @@ void main() {
 
     req = new Query<TestModel>()
       ..predicate = new QueryPredicate("id = @id", {"id": item.id})
-      ..propertiesToFetch = ["id", "name"];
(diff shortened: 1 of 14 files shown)


test/db/postgresql/fetch_test.dart
```diff
@@ -162,8 +153,8 @@ void main() {
     var req = new Query<TestModel>()..values = m;
     await req.insert();
 
-    req = new Query<TestModel>();
-    req.propertiesToFetch = ["id", "badkey"];
+    req = new Query<TestModel>()
+      ..returningProperties((t) => [t.id, t["badkey"]]);
```
MarcusSmith: I like how intentional you have to be to fetch a bad key now

PR #205, opened 2017-02-17
Allows joinOne/joinMany to be used with pageBy

joeconwaystk:
Rebase against #204.


joeconwaystk:
Well this doesn't work

PR #210, opened 2017-02-22
Allows AuthServer to manage scope. 

joeconwaystk:
Implements scope storage in managed_auth.

### DIFF
--- a/test/auth/auth_scope_test.dart
+++ b/test/auth/auth_scope_test.dart
@@ -171,11 +171,45 @@ void main() {
 
   group("Client behavior", () {
     test("Client collapses redundant scope because of nesting", () {
-      fail("NYI");
+      var c = new AuthClient("a", "b", "c", allowedScopes: [
+        new AuthScope("abc"),
+        new AuthScope("abc:def")
+      ]);
+      expect(c.allowedScopes.length, 1);
+      expect(c.allowedScopes.first.isExactly("abc"), true);
+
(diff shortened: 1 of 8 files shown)


test/auth/auth_scope_test.dart
```diff
@@ -137,4 +137,79 @@ void main() {
...
     expect(scope.allows("user:location:equipment"), false);
   });
+
+  test("Can contain all valid characters", () {
+    var scope = new AuthScope("ABC:DEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyzz0123456789!#\$%&'`()*+,./;<=>?@[]^_{|}-");
+    expect(scope.allows("ABC:DEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyzz0123456789!#\$%&'`()*+,./;<=>?@[]^_{|}-"), true);
+    expect(scope.allows("ABC"), true);
+  });
+
+  test("Cannot contain invalid characters", () {
```
MarcusSmith: Since scopes are stored joined by " " I think it would be good to add a test to specifically ensure that a " " can't be used inside a scope

joeconwaystk: Technically, it'd be two different scopes then

joeconwaystk: oh i see what you are saying - ok got it

PR #218, opened 2017-03-10
Added handling for inserting empty object

anachlas:


lib/src/db/postgresql/postgresql_query.dart
```diff
@@ -21,8 +21,16 @@ class PostgresQuery<InstanceType extends ManagedObject> extends Object
 
     var buffer = new StringBuffer();
     buffer.write("INSERT INTO ${builder.primaryTableDefinition} ");
-    buffer.write("(${builder.valuesColumnString}) ");
-    buffer.write("VALUES (${builder.insertionValueString}) ");
+
+    if (builder.valuesColumnString.isNotEmpty) {
```
joeconwaystk: Is there another way to check these for emptiness? I think the string builder getters on this type do a bit of work and the info they are derived from may be a better place to check

anachlas: Yes I can use the `columnValueMappers`, good call!

PR #221, opened 2017-03-13
Change default port

joeconwaystk:
Defaults to 8081 to avoid conflict with pub serve for development in parallel 
Allow system assigned ports to run tests in parallel 
Reset logger listeners during tests so they stop accruing newlines


joeconwaystk:
Oh yeah, good call, yeah I think so. I'll add that and I apparently need to add some code coverage for that new constructor

PR #232, opened 2017-04-10
Print CLI and project version when running tools

joeconwaystk:
Also removes unnecessary bash script from project


joeconwaystk:
Yeah, I think there is a much safer way of doing this, and I'm realizing now there is potential for a problem if the project and CLI are mismatched. Gonna close for now and reopen when I have a better plan

PR #235, opened 2017-04-21
Fixing isolate hang introduced by 1.23

joeconwaystk:
Hopefully resolves it.


joeconwaystk:
Dropped an entire percentage point of code coverage, since the ending curly bracket now counts as a 'miss'. So this was a drop, but a meaningless one.

PR #267, opened 2017-05-30
Better names for some properties, deprecating ones that were actually used

joeconwaystk:


test/base/isolate/isolate_failure_test.dart
```diff
@@ -120,15 +120,17 @@ class TimeoutSink extends RequestSink {
...
+    var elapsed = 0;
+    var timer = new Timer.periodic(new Duration(milliseconds: 500), (t) {
+      elapsed += 500;
+      print("waiting...");
+      if (elapsed > timeoutLength * 1000) {
+        completer.complete();
+        t.cancel();
+      }
+    });
+    await completer.future;
```
anachlas: So why did the old code fail sometimes?

joeconwaystk: Something to do with Future.delayed and perhaps the length of it? Some things turned up scouring the internet about Travis-Ci stalls and time outs seemed to be the cause.. so this is basically just another timeout mechanism but doesn't go too long without processing something.

PR #271, opened 2017-06-08
Adds ApplicationMessageHub to send cross-isolate messages.

joeconwaystk:


lib/src/application/application.dart
```diff
@@ -113,6 +124,8 @@ class Application<RequestSinkType extends RequestSink> {
         logger.severe("$e", this, st);
         rethrow;
       }
+      _hasFinishedLaunching = true;
+      supervisors.forEach((sup) => sup.sendPendingMessages());
```
erikist: `sup` dawg

PR #288, opened 2017-06-28
Adding a "few" more guides

joeconwaystk:


source/docs/testing/database.md
```diff
@@ -0,0 +1,80 @@
...
+
+Use [configuration files](../http/configure.md) to manage which database an application connects to. This may or may not be checked into source control, depending on a team's preference. Control which file is loaded with command-line options to `aqueduct serve` or the `bin/main.dart` script:
+
+```
+aqueduct serve -c local.yaml
+```
+
+## Have Scripts to Provision Based on Scenarios
+
+It is often the case that you will want to have a certain set of data in an local database for the purpose of testing a client application. Create `bin` scripts to provision the database and add the desired data. For example, you might have a script named `bin/ios_integration.dart` that re-provision a database and inserts data into it using `Query<T>` instances and the `ManagedObject<T>`s declared in your application.
```
anachlas: re-provisions

source/docs/testing/mock.md
```diff
@@ -0,0 +1,107 @@
...
+  var response =
+    await app.client.authenticatedRequest("/github_profile/fred").get();
+
+  var requestSentByYourApplicationToGitHub = await githubMock.next();
+  expect(requestSentByYourApplicationToGitHub.method, "GET");
+  expect(requestSentByYourApplicationToGitHub.path, "/users/search?name=fred");
+});
+```
+
+In the above code, we expecting that anytime the request `GET /github_profile/fred` is sent to your application, that it turns around and searches for a user in GitHub's API. This test ensures that we have correctly translated our request to a request to be made to the GitHub API. If no request was made - because of a programmer error - this test would fail because the `Future` returned from `githubMock.next()` would never complete. There is no next request, because none was ever delivered!
```
anachlas: we are expecting

source/docs/testing/tests.md
```diff
@@ -0,0 +1,374 @@
...
+test("That we get a 200 from /endpoint", () async {
+  var response = await app.client.request("/endpoint").get();
+
+  expect(response, hasStatus(200));
+});
+```
+
+## Using a TestClient
+
+A `TestClient` creates requests (instances of `TestRequest`), which have execution methods (like `get` and `post`) that returns responses (instances of `TestResponse`). The purpose of an Aqueduct test is to ensure that a request elicits the intended response. For example, you may want to make sure that a request with all the right parameters returns a response with the expected status code and JSON response body. Likewise, you may want to ensure that a request with some invalid parameters returns a response with the appropriate error information.
```
anachlas: that return responses

source/docs/testing/tests.md
```diff
@@ -0,0 +1,374 @@
...
+
+Once a `TestRequest` is executed and returns a `TestResponse`, the real work begins: verifying the response is what you expect. A `TestResponse` has properties for the things you would typically expect of an HTTP response: status code, headers and body. In addition to the raw string `body` property, the following body-inspecting properties exist:
+
+- `decodedBody` is the object created by decoding the response body according to its content-type
+- `asList` is `decodedBody`, but cast to a `List`
+- `asMap` is `decodedBody`, but cast to a `Map`
+
+The great part about each of these three methods is that if the body cannot be decoded according to its content-type, or cannot be cast into the expected type, an exception is thrown and your tests fail. In other words, these methods implicitly test the validity of the response body.
+
+Using individual properties of a `TestResponse` in test expectations is a valid use case, but there are some more helpful utilities for verifying response that create more test code.
```
anachlas: How about, verifying responses that create clearer test code.?

source/docs/testing/tests.md
```diff
@@ -0,0 +1,374 @@
...
+
+  await createDatabaseSchema(ManagedContext.defaultContext);
+
+  client = new TestClient(application);
+}
+```
+
+Notice that the `ManagedContext.defaultContext` will have already been set by the application's `RequestSink`.
+
+After a test is executed, the test database should be cleared of data so that none of the stored data test leaks into the next test. Because starting and stopping an application isn't a cheap operation, it is often better to simply delete the contents of the database rather than restart the whole application. This is why the flag `isTemporary` in `SchemaBuilder.toSchema` matters: it creates *temporary* tables that only live as long as the database connection. By simply reconnecting to the database, all of the tables and data created are discard. Therefore, all you have to do is close the connection and add the database schema again.
```
anachlas: discarded

source/docs/tour.md
```diff
@@ -0,0 +1,422 @@
...
+
+  @ManagedColumnAttributes(indexed: true)
+  String name;
+
+  DateTime hireDate;
+  int salary;
+}
+```
+
+Each property in the private class
```
anachlas: Sentence fragment

joeconwaystk:
@anachlas Danke, feedback addressed

PR #289, opened 2017-06-28
Adds behavior to allow overriding persistent properties

joeconwaystk:


anachlas:
Even better 👍

joeconwaystk:
Value of code coverage - that whole thing never actually ran the two lines of code to swap out two matching ivars

PR #290, opened 2017-06-29
Cleanup docs for initializeApplication and remove from templates

joeconwaystk:


example/templates/db/lib/wildfire_sink.dart
```diff
@@ -35,19 +35,6 @@ class WildfireSink extends RequestSink {
...
   }
 
-  /// Do one-time application setup in this method.
-  ///
-  /// This method is executed before any instances of this type are created and is the first step in the initialization process.
-  ///
-  /// Values can be added to [appConfig]'s [ApplicationConfiguration.options] and will be available in each instance of this class
-  /// in the constructor.
-  static Future initializeApplication(ApplicationConfiguration appConfig) async {
-    if (appConfig.configurationFilePath == null) {
```
MarcusSmith: Is this check being handled somewhere else already, or was it just not needed?

joeconwaystk: It wasn't really needed IMO

PR #301, opened 2017-07-13
Add arbitrary HTTP method for test client.

terrencepreilly:
Adds an orthogonal method, `method(String verb)` to test client, to match HTTP Controllers such as `@HTTPMethod(String verb)`.  I was testing an endpoint with PATCH and found that only the main verbs were supported.  It would be nice to be able to test any verbs.


joeconwaystk:
👍 Will merge when tests complete. Good addition.

PR #310, opened 2017-07-21
Add addResponseModifier docs

joeconwaystk:


source/docs/http/request_controller.md
```diff
@@ -143,6 +143,24 @@ Future<RequestOrResponse> processRequest(Request request) async {
 }
 ```
 
+## Middleware Modifying a Response
+
+A `RequestController` that doesn't respond to a request can still modify the eventual response. This is valuable for "middleware" `RequestController` that have some information to return back to the client, but aren't responsible for generating the response. For example, the following shows a request channel where `UserController` will create the response, but middleware will add a header to the response:
```
anachlas: valuable for "middleware" `RequestController`s

PR #312, opened 2017-07-25
Allows customizable max entity size

joeconwaystk:
@MarcusSmith Let's sync up to test of your stuff against this branch


MarcusSmith:
This is working for me.

Could we make sure that the max entity size is listed in the docs somewhere and that we have instructions for how / when is best to set it if you want something other than the default?

PR #316, opened 2017-08-08
Adds multi-column unique to Schema and schema diff

joeconwaystk:
PR 2/3 (rebase off #314) for multi-column unique. Really need a better name for this.

This adds multi-column unique to SchemaTable; two SchemaTables are not equivalent if they have different values for this property.


erikist:
Looks good to me

PR #323, opened 2017-09-13
Adds SIGINT handler to shut down app gracefully

joeconwaystk:


MarcusSmith:
Is there a way to force a SIGINT in a unit test to ensure that it stops the app and throws the correct exception?

joeconwaystk:
I think technically just invoking the interrupt handler would do it, but I don't want to make it part of the public API. I'll think on that some, tho.

anachlas:
Something in `isolate_failure_test.dart` keeps hanging on Travis

joeconwaystk:
Yeah those tests are the ones that stall sometimes on travis will re run

PR #326, opened 2017-09-19
Note about expiration time

erikist:
Just added that note I mentioned beforee

### DIFF
--- a/source/docs/auth/controllers.md
+++ b/source/docs/auth/controllers.md
@@ -58,7 +58,8 @@ The response to a password token request is a JSON body that follows the OAuth 2
 ```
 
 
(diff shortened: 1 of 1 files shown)


source/docs/auth/controllers.md
```diff
@@ -57,6 +57,8 @@ The response to a password token request is a JSON body that follows the OAuth 2
 }
 ```
 
+> Note: the `expires_in` field is a computed property based on the delta of the issue date and expiration date. Take care manually editing the database.
```
joeconwaystk: There is a special note syntax for the markdown. Here's an example:

```
!!! warning "Closing the Application"
    Once you're done running an application, stop it with `^C`. Otherwise, the next time you try and start an application, it will fail because your previous application is already listening for requests on the same port.
```

PR #400, opened 2018-03-09
Minor clean up in the http docs

MarcusSmith:


source/docs/http/resource_controller.md
```diff
@@ -8,7 +8,7 @@ A `ResourceController` is a [controller](controller.md) that provide convenience
 - updating an individual user (`PUT /users/:id`)
 - deleting an individual user (`DELETE /users/:id`)
 
-A `ResourceController` subclass implements an *operation method* for each operation that it supports.
+A `ResourceController` subclass implements at least one *operation method* for each operation that it supports.
```
joeconwaystk: This would lead me to believe that you could implement more than one operation method for an operation?

source/docs/http/websockets.md
```diff
@@ -75,7 +75,7 @@ router
 If we continue with the 'chat application' example, the code for `handleEvent` may be something like:
 
 ```dart
-void handleRequest(dynamic event, {int fromUserID}) {
+void handleEvent(dynamic event, {int fromUserID}) {
```
joeconwaystk: 👍

source/docs/http/resource_controller.md
MarcusSmith: You could have a GET with no ID and a GET with an ID, right?

joeconwaystk: Yes, but an operation is defined as the method + path. So GET /path and GET /path/id are two separate ops.

MarcusSmith: 👍

PR #402, opened 2018-03-12
Expose ManagedRelationshipType

sudolibre:
ManagedRelationshipType is exposed via the relationshipType property on ManagedRelationshipDescription however the type itself is not exposed outside the package. This means that we cannot perform checks such as
`if (managedRelationshipDescription.relationshipType == ManagedRelationshipType.hasMany)`

This PR is to expose this type to enable such checks.

### DIFF
--- a/lib/aqueduct.dart
+++ b/lib/aqueduct.dart
@@ -47,4 +47,4 @@ export 'src/auth/auth.dart';
 export 'src/db/db.dart';
 export 'src/http/http.dart';
 export 'src/openapi/openapi.dart';
-
+export 'src/db/managed/relationship_type.dart';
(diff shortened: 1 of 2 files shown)


joeconwaystk:
Can you import 'package:aqueduct/src/db/managed/relationship_type.dart' from your app without exporting this?

sudolibre:
> Can you import 'package:aqueduct/src/db/managed/relationship_type.dart' from your app without exporting this?

I can but it causes this warning from the analysis server: ```Don't import implementation files from another package. (implementation_imports at [reservoir] lib/src/entity_controller.dart:5)```

joeconwaystk:
Fair enough

joeconwaystk:
Can you add documentation comments to the enum if it's going to be public facing?

sudolibre:
> Can you add documentation comments to the enum if it's going to be public facing?
> 
> View changes

added the following comment: ```/// The possible database relationships.```

PR #403, opened 2018-03-14
Allows for more convenient building of insertion/update values in Query

joeconwaystk:
- Prevents user from setting has-* relationships in Query.values.
- Accessing belongs-to relationships automatically creates an empty instance if object is Query.values.
- Prevents user from setting values other than foreign key for belongs-to relationships in Query.values
- Allows user to set Query.values from an object created thru default constructor, and strips away invalid insertion values (has-many, has-one, etc.)

### DIFF
--- a/lib/src/db/query/query.dart
+++ b/lib/src/db/query/query.dart
@@ -239,6 +239,9 @@ abstract class Query<InstanceType extends ManagedObject> {
   ///           ..values.name = "Sally"
   ///           ..values.manager.id = 10;
   ///         await q.insert();
+  ///
+  /// You may replace this property with a new instance of [InstanceType]. When doing so, the provided instance's [ManagedObject.backing]
+  /// is altered and any properties that are not allowed to be set on [values] are stripped away.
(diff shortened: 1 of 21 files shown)


lib/src/db/query/query.dart
```diff
@@ -202,20 +202,46 @@ abstract class Query<InstanceType extends ManagedObject> {
...
-  /// Values to be used when inserting or updating an object.
+  /// Values to be sent to database during an [update] or [insert] query.
+  ///
+  /// You set values for the properties of this object to insert a row or update one or more rows.
+  /// This property is the same type as the type being inserted or updated. [values] is empty (but not null)
+  /// when a [Query] is first created, therefore, you do not have to assign an instance to it and may set
+  /// values for its properties immediately:
+  ///
+  ///         var q = Query<User>()
+  ///           ..values.name = 'Joe
```
anachlas: Unclosed quote

joeconwaystk: danke

PR #406, opened 2018-03-15
OneOf Validations for DateTime

sudolibre:
OneOf validation should only be used for strings and ints.

~~Previously Managed Validator would throw an error if you attempted to do a oneOf validation on a DateTime property. This is because the options coming from the annotations were of type String and the oneOf logic checks to ensure the type of all the options are equal to or a subtype of the property type e.g. (String is DateTime).~~

~~The solution is to store the options as DateTime when the attribute type is DateTime~~


joeconwaystk:
Can DateTime not be const? e.g. `@Validate.oneOf(const [const DateTime(2016)])`

sudolibre:
> Can DateTime not be const? e.g. @Validate.oneOf(const [const DateTime(2016)])

It looks like no. It gives a message saying the constructor is not a const constructor. None of the constructors in the documentation are marked as const either.

joeconwaystk:
Is there a reason to enable this, btw? I'm not able to think of a use case where a field would have a finite set of dates.

sudolibre:
> Is there a reason to enable this, btw? I'm not able to think of a use case where a field would have a finite set of dates.

Mainly for correctness. I don't have a use case in mind where it would be static like that but the API allows it.

joeconwaystk:
Does it make more sense to throw a data model error when setting a oneOf validator on an unsupported field type? Other than String or int, I can't think of another option

sudolibre:
> Does it make more sense to throw a data model error when setting a oneOf validator on an unsupported field type? Other than String or int, I can't think of another option

I think thats more of a philosophical question. If you want to to take a YAGNI approach and restrict the use cases to the necessary or obvious then throwing an error makes more sense. When we abstract the validator logic out we can have an ORM layer that adds these restrictions while keeping it more generic for other use cases. If the inputs for the validators didn't have const requirement and it could be a dynamic list I can see how there may be more uses cases for something like this.

joeconwaystk:
The const requirements are a requirement of an annotation, so we can't go down that path. In general, I'd like to restrict stuff like this statically, but Dart's generics just aren't powerful enough to do that, so we need runtime errors. IMO it's better to throw an error than add support for something that probably doesn't have a real world use case. Esp. because this means we have to support it going forward, making change more costly

lib/src/db/managed/validate.dart
```diff
@@ -105,6 +105,16 @@ class ManagedValidator {
       _buildLengthExpressions();
       _validationMethod = _validateExpressions;
     } else if (definition._builtinValidate == _BuiltinValidate.oneOf) {
+      var supportedOneOfTypes = [
+        ManagedPropertyType.string,
+        ManagedPropertyType.integer,
+        ManagedPropertyType.doublePrecision,
```
joeconwaystk: I'd argue that double shouldn't be supported, given that it's hard to exactly represent a floating point number, and this could lead to some confusion

joeconwaystk:
Superb. Will merge when tests finish

PR #424, opened 2018-04-25
updating git clone link

shomodj:
when using ssh+git users will get permission denied, so switched to https


joeconwaystk:
👍 Will deploy to site later today, thank you.

PR #489, opened 2018-07-06
Cleans up remaining one-off implicit casts

joeconwaystk:
Left in implicit casts for headers and query parameters - those deserve additional attention.


joeconwaystk:
This needs to be merged with 3.0 once #487 completes

Issue #527, opened 2018-08-07
Controller should default to throwing an error at startup if it doesn't not override handle


joeconwaystk:
If someone makes a mistake by not overriding `handle`, they will get no indicator that the request was not handled. 

- Some opaque controllers (e.g., RouteController) rely on the default behavior passing the request to linked controller.

PR #535, opened 2018-08-13
Controller is now abstract.

joeconwaystk:
Was previously a concrete class where you a user had to override `handle`, but it is preferable to require an implementation for this method when directly subclassing. This is a minor breaking change that should have no real impact.


joeconwaystk:
Finishes #527.

Issue #459, opened 2018-06-11
Add optional values argument to Query constructor


joeconwaystk:
For queries that will insert/update rows, an optional argument with those values is a useful shortcut that can infer the `T` of `Query<T>`. 

Usage:

```dart
final thing = new Thing(...);
final q = new Query(context, values: thing)
  ..where((t) => t.id).equalTo(1); // is inferred to Query<Thing>
await q.update();
```

jhdoak:
@joeconwaystk - Can this issue be closed with the merge of #558?

joeconwaystk:
Oh yup, thank you!

PR #558, opened 2018-09-26
Adds optional values argument to Query constructor

joeconwaystk:
Allows for Query parameterized type to be inferred from this argument, e.g. `Query(context, MyObject())`.


joeconwaystk:
Finishes #459

PR #562, opened 2018-10-08
Refactors data model building

joeconwaystk:
Also adds `name` to `@Table` definition and deprecates `tableName` static method.


aqueduct/lib/src/db/managed/data_model_builder.dart
```diff
@@ -1,484 +1,37 @@
...
-      entity.relationships.forEach((_, rel) {
-        if (rel.relationshipType == ManagedRelationshipType.belongsTo) {
-          var foreignKey = rel.destinationEntity.relationships?.values
-              ?.firstWhere(
-                  (r) =>
-                      r.relationshipType == ManagedRelationshipType.belongsTo &&
-                      r.destinationEntity == entity,
-                  orElse: () => null);
-          if (foreignKey != null) {
-            throw ManagedDataModelError.cyclicReference(entity,
```
anachlas: Does this need to be checked for somewhere?

joeconwaystk: It does - that failing test is due to this. I also want to allow cyclic references in the future. I'm debating on deleting the test case versus re-implementing this check to keep master clean.

anachlas: However cyclic references are handled there will probably still need to be code to detect them, I would say keep the check in there for now but its up to you.

Issue #514, opened 2018-07-23
Add @mustCallSuper where appropriate


halldors:
Many places in the code there are comments like these. "If you override this method, you *must* invoke the super implementation"

Add `@mustCallSuper` to those methods so that tools like IntelliJ and VS Code will warn developers when they miss it.

PR #564, opened 2018-10-10
Adding Must Call Super Annotations

sudolibre:
Issue #514 

Searched for "superclass" and "super" with a space before or after to find comments stating the method must call super when overridden.

This caused warnings to popup for Controller.didAddToChannel overrides. Specifically for the Router and _ControllerRecycler class.
screen shot 2018-10-10 at 10 57 35 am

### DIFF
--- a/aqueduct/lib/src/application/channel.dart
+++ b/aqueduct/lib/src/application/channel.dart
@@ -177,6 +177,7 @@ abstract class ApplicationChannel implements APIComponentDocumenter {
     return doc;
   }
 
+  @mustCallSuper
   @override
   void documentComponents(APIDocumentContext registry) {
     entryPoint.documentComponents(registry);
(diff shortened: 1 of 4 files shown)


aqueduct/lib/src/application/channel.dart
```diff
@@ -177,6 +177,7 @@ abstract class ApplicationChannel implements APIComponentDocumenter {
     return doc;
   }
 
+  @mustCallSuper
```
joeconwaystk: What was the intent of requiring super here? Its technically not required, but I could maybe see a reason not to in some dire circumstances? Was it causing an issue anywhere?

sudolibre: If you read the description of the API above it mentions: 
```
If there services that are part of the application, but not stored as properties of this channel, you may override
  /// [documentComponents] in your subclass to add them. You must call the superclass' implementation of [documentComponents].
```

joeconwaystk: Gotcha. OK, I'm fine with keeping it.

PR #568, opened 2018-10-10
Add implicit flow to OpenAPI documentation

jhdoak:


aqueduct/lib/src/auth/auth_redirect_controller.dart
```diff
@@ -216,8 +216,9 @@ class AuthRedirectController extends ResourceController {
   Map<String, APIOperation> documentOperations(
       APIDocumentContext context, String route, APIPath path) {
     final ops = super.documentOperations(context, route, path);
-    authServer.documentedAuthorizationCodeFlow.authorizationURL =
-        Uri(path: route.substring(1));
+    final urlPath = route.substring(1);
```
joeconwaystk: Uri(path: route.substring(1)) here, and then use that var below - no need to instantiated to Uris

PR #569, opened 2018-10-10
Add Implicit Flow to AuthRedirectController

MarcusSmith:
This is untested, but wanted to go ahead and get eyes on the code

### DIFF
--- a/aqueduct/lib/src/auth/auth_redirect_controller.dart
+++ b/aqueduct/lib/src/auth/auth_redirect_controller.dart
@@ -283,13 +283,7 @@ class AuthRedirectController extends ResourceController {
         queryParameters["error"] = error.reasonString;
       }
     } else if (responseType == "token") {
-      Map<String, dynamic> params;
-
-      if (token != null) {
-        params = token.asMap();
-      } else {
-        params = {};
-      }
+      final params = token?.asMap() ?? {};
 
       if (clientStateOrNull != null) {
         params["state"] = clientStateOrNull;
--- a/aqueduct/lib/src/auth/auth_redirect_controller.dart
+++ b/aqueduct/lib/src/auth/auth_redirect_controller.dart
@@ -94,6 +94,10 @@ class AuthRedirectController extends ResourceController {
       return Response(405, {}, null);
     }
 
+    if (!_isValidResponseType(responseType)) {
+      return _unsupportedResponseTypeResponse;
+    }
+
     final renderedPage = await delegate.render(
         this, request.raw.uri, responseType, clientID, state, scope);
     if (renderedPage == null) {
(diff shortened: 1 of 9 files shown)


aqueduct/lib/src/auth/auth_redirect_controller.dart
```diff
@@ -233,15 +242,37 @@ class AuthRedirectController extends ResourceController {
     final redirectURI = Uri.parse(uriString);
     final queryParameters =
```
sudolibre: I know this isn't part of your changes but Uri.parse can throw. Shouldn't it return a bad response for this?

joeconwaystk: Yes. It'd yield a 500 right now. It should probably yield the same result as if the uriString were null

aqueduct/lib/src/auth/auth_redirect_controller.dart
```diff
@@ -52,6 +52,8 @@ class AuthRedirectController extends ResourceController {
     ];
   }
 
+  static Response _unsupportedResponseTypeResponse = Response.badRequest(body: "<h1>Error</h1><p>unsupported_response_type</p>")..contentType = ContentType.html;
```
jhdoak: It might be nice to find a way to include the unsupported response type in this response. `Unsupported response type foo`.

aqueduct/lib/src/auth/auth_redirect_controller.dart
```diff
@@ -222,26 +231,55 @@ class AuthRedirectController extends ResourceController {
...
+      if (clientStateOrNull != null) {
+        queryParameters["state"] = clientStateOrNull;
+      }
+      if (error != null) {
+        queryParameters["error"] = error.reasonString;
+      }
+    } else if (responseType == "token") {
+      Map<String, dynamic> params;
+
+      if (token != null) {
```
jhdoak: I think this could be written as a ternary along with the declaration, but that's a matter of preference. You know how  much I love me a ternary.

aqueduct/lib/src/auth/auth_redirect_controller.dart
MarcusSmith: `unsupported_response_type` is in the RFC itself. We are giving all the other errors without additional context, and this is the easiest one to understand, I'd rather keep these consistent

aqueduct/lib/src/auth/auth_redirect_controller.dart
MarcusSmith: This is the perfect place for it :)

MarcusSmith: Nevermind, Dart has optional chaining, way better than ternary

PR #570, opened 2018-10-10
[DO NOT MERGE: WIP] Insert Iterable

sudolibre:
Brute force implementation but tests should pass.

This adds a valuesList property to Query and a insertList() method to handle collections.

@joeconwaystk can you give this a look when you get a chance and see if this design is worth cleaning up, if we want to do a larger refactor, or if i'm missing the mark here. I think it makes sense that values and insert both were List<InstanceType> but that would be a breaking change.

### DIFF
--- a/aqueduct/lib/src/db/postgresql/postgresql_query.dart
+++ b/aqueduct/lib/src/db/postgresql/postgresql_query.dart
@@ -30,6 +30,12 @@ class PostgresQuery<InstanceType extends ManagedObject> extends Object
 
   @override
   Future<InstanceType> insert() async {
+    final list = await insertList();
+    return Future.value(list.first);
+  }
+
+  @override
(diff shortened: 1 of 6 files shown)


aqueduct/lib/src/db/postgresql/postgresql_query.dart
```diff
@@ -30,18 +30,24 @@ class PostgresQuery<InstanceType extends ManagedObject> extends Object
 
   @override
   Future<InstanceType> insert() async {
+    final list = await insertList();
+    return Future.value(list.first);
+  }
+
+  @override
+  Future<List<InstanceType>> insertList() async {
```
joeconwaystk: This method should still just always be `insert` - each of the execution commands (fetch, update, select, delete) are one-to-one with a SQL operation

PR #572, opened 2018-10-11
Add static method on Query for inserting multiple objects

sudolibre:


aqueduct/lib/src/db/query/query.dart
```diff
@@ -60,6 +60,14 @@ abstract class Query<InstanceType extends ManagedObject> {
     return query.insert();
   }
 
+  /// Inserts [object]s into the database managed by [context].
```
jhdoak: Pulled this down, and IntelliJ is telling me `[objects]` would be preferred to `[object]s` here, to match up with the method's parameter name.

aqueduct/lib/src/db/query/query.dart
```diff
@@ -60,6 +60,14 @@ abstract class Query<InstanceType extends ManagedObject> {
     return query.insert();
   }
 
+  /// Inserts [objects] into the database managed by [context].
+  ///
+  /// This currently has no Query instance equivalent
```
joeconwaystk: Can you describe that it inserts each object in objects within a single transaction?

sudolibre: I changed this line idk why it's still showing the old version on the main page of this PR

PR #575, opened 2018-10-12
Auth redirect controller

MarcusSmith:


aqueduct/lib/src/auth/auth_redirect_controller.dart
```diff
@@ -0,0 +1,317 @@
...
+    ];
+  }
+
+  static Response _unsupportedResponseTypeResponse = Response.badRequest(body: "<h1>Error</h1><p>unsupported_response_type</p>")..contentType = ContentType.html;
+
+  /// A reference to the [AuthServer] used to grant authorization codes and access tokens.
+  final AuthServer authServer;
+
+  /// When true, the controller allows for the Implicit Grant Flow
+  bool allowsImplicit;
```
joeconwaystk: final

aqueduct/lib/src/auth/auth_redirect_controller.dart
```diff
@@ -0,0 +1,317 @@
...
+            contentTypes: ["text/html"])
+      };
+    } else if (operation.method == "POST") {
+      return {
+        "${HttpStatus.movedTemporarily}": APIResponse(
+            "If successful, in the case of a 'response type' of 'code', the query "
+            "parameter of the redirect URI named 'code' contains authorization code. "
+            "Otherwise, the query parameter 'error' is present and contains a error string. "
+            "In the case of a 'response type' of 'token', the redirect URI's fragment "
+            "contain san access token. Otherwise, the fragment contains an error code.",
```
joeconwaystk: contain san -> contains an

aqueduct/lib/src/auth/auth_redirect_controller.dart
```diff
@@ -0,0 +1,317 @@
...
+      return _unsupportedResponseTypeResponse;
+    }
+
+    final responseURI = Uri(
+        scheme: redirectURI.scheme,
+        userInfo: redirectURI.userInfo,
+        host: redirectURI.host,
+        port: redirectURI.port,
+        path: redirectURI.path,
+        queryParameters: queryParameters,
```
joeconwaystk: Is it intended that the implicit flow would include query parameters if the redirect uri had query parameters?

aqueduct/lib/src/cli/commands/auth_add_client.dart
```diff
@@ -78,14 +78,6 @@ class CLIAuthAddClient extends CLICommand
       return 1;
     }
 
-    if (secret == null && redirectUri != null) {
```
joeconwaystk: May want to double-check the Auth CLI documentation guide to make sure this isn't referenced.

aqueduct/test/auth/auth_redirect_controller_test.dart
```diff
@@ -0,0 +1,1008 @@
+import 'dart:async';
```
joeconwaystk: Since this file was moved, I can't see what changed - is there some sort of marker where tests started changing? Did any of the existing tests change?

aqueduct/lib/src/auth/auth_redirect_controller.dart
MarcusSmith: Yes: https://tools.ietf.org/html/rfc6749#section-3.1.2

aqueduct/test/auth/auth_redirect_controller_test.dart
MarcusSmith: Where I could I made new test groups, prefixing the old group title with `Code` and the new group with related tests with `Token`. Tweaked some existing tests, allowing public clients with redirect URIs. There are a few tests where it previously redirected back to the client's URI that now return 400s with the basic HTML error message

aqueduct/lib/src/auth/auth_redirect_controller.dart
MarcusSmith: 😉 👉

MarcusSmith:
It should be added back, but I don't think I have it deprecated, I'll give that a look and add that in

PR #588, opened 2018-11-14
Improves test code for migration

joeconwaystk:
This is a refactor of the code for testing migrations in an effort to speed up execution time and make test scenarios easier to write in the future. This included a refactor of how migrations are generated. `SchemaBuilder` previously only generated SQL commands when invoking methods like `createTable`. These commands are specific to the SQL-flavor of the database the migration is being executed on. If you do not specify a database to this command, it now generates Dart code that replicates the code invoked on it. For example, if you invoke `builder.createTable(...)`, it will generate the statement `builder.createTable(...)`.


joeconwaystk:
This also changes the default behavior of generated migrations. Previously, if you generated a migration in which you altered a table that had existing rows to have a column that is nullable, `db generate` would intentionally generate an invalid migration file that required you to select a default value for that column. This is no longer the case, instead, you will get a warning when you generate and an error when you try to migrate. In doing so, the user is no longer forced to provide an initial value for the column if it is not necessary (i.e. there are no rows in the table0.

Issue #447, opened 2018-06-10
Provide better control over ORM logging


joeconwaystk:
A user should be able to indicate which log level will emit a log record for a query. The query should log its format string and its parameters if a log listener is attached for the indicated log level. Parameters should be able to be removed from the log record if they are sensitive. These options might make sense as settings for a persistent store.

PR #591, opened 2018-11-15
Log warning if PostgreSQL throws an exception.

lieblb:
A small issue I came across today: with the default `INFO` log level, PostgreSQL exceptions are not logged, which means an insert can fail silently without any output, which is quite confusing.

In my concrete case, I had a null constraint violation that prevented the insertion from happening. With the PR I now get a proper error message:

`[WARNING] aqueduct: PostgreSQLSeverity.error 23502: null value in column "password" violates not-null constraint Detail: Failing row contains (13, Mein toller Test, {"version": "1.0", "encoding": "utf-8", "questestinterop": {"ass..., 2018-11-15 11:49:53.67694, 0, null, 0). Table: _exam Column: password`

Not sure if it should be warning or an error, and not sure how to really log this. Anyway, this PR pins the location where something should happen I guess.


joeconwaystk:
Thanks. I'm good with this. There is an open issue #447 to look at this more deeply, but I think this is helpful in the interim.

Issue #604, opened 2018-12-18
3.1 Regression: Cannot read foreign key from request body


joeconwaystk:
Ignoring auto-incrementing fields when reading a managed object with readFromMap was added in 3.1. The purpose of this behavior is to prevent an API user from overwriting an auto-incrementing primary key if the developer failed to strip that key from the payload when the server received it.

This had the unintended effect of preventing foreign key reads through the same mechanism if the related primary key is auto-incrementing. Likewise, if a user POSTs a collection of objects with object identifiers, the expected behavior would be to read those identifiers and not ignore them.

joeconwaystk:
The proposed solution is to add an optional boolean flag to both Bind.body and readFromMap for this behavior. To keep backwards compatibility, this flag should default to 3.0 behavior.

joeconwaystk:
Another solution is to allow the developer to customize ignored, allowed, or 'error' keys. e.g. `readFromMap(map, ignore: [#id], errorIfExist: [#name])` and same for Bind.body.

PR #609, opened 2018-12-29
Adds behavior for filtering/validating keys of request body objects

itsjoeconway:
e.g. 

Used in body bindings:
`@Bind.body(ignore: ["id"], require: ["name"], reject: ["password"])`

And as a replacement to readFromMap:
`object.read(body, ignore: ["id"], require: ["name"], reject: ["password"]);`


joeconwaystk:
Finishes #604

PR #622, opened 2019-01-18
Adds ability to remove multiple properties from backing map in a single call

tgercek:
Example:

`userProfile.removePropertiesFromBackingMap(["createdAt", "updatedAt"]);`

### DIFF
--- a/aqueduct/test/db/model_test.dart
+++ b/aqueduct/test/db/model_test.dart
@@ -414,6 +414,71 @@ void main() {
     expect(m["bothOverQualified"], "foo");
   });
 
+  test("Can remove single property from backing map", () {
+    var m = (User()
+      ..id = 1
+      ..name = "Bob"
+      ..dateCreated = DateTime(2018,1,30))
+      ..removePropertyFromBackingMap("name")
+      ..asMap();
+
+    expect(m["id"], 1);
+    expect(m["name"], isNull);
+    expect(m["dateCreated"], DateTime(2018,1,30));
+  });
+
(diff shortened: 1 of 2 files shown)


joeconwaystk:
This is fine to add as long as it has tests. Also, a feature for a similar use case was added in #609 that you might be interested in. 

For tests, add to `tests/model_test.dart`: 1. ensure call with 1 key removes just that key. 2. ensure call with 2 keys removes both of those keys. 3. ensure call where none of the keys listed are in the object has no impact on the object.

Btw, this prompted me to check and there is no direct coverage for the singular method. If you can add a similar set of tests for the singular method, that'd be a great bonus.

aqueduct/test/db/model_test.dart
```diff
@@ -414,6 +414,71 @@ void main() {
     expect(m["bothOverQualified"], "foo");
   });
```
joeconwaystk: Tests look nice and clean. Two things to make sure we get the right behavior. First, `m` is always an instance of `User` and the `asMap` is actually discarded. The fact that `m` can be accessed like a map (i.e. `[key]`) is because so can all managed objects. You'll have to assign the user to a temporary variable and then store the result of `asMap` in `m`. Then, use `containsKey` instead of checking for null to protect against the case of the key existing, but the value being null for which this test would successfully trigger if true.

tgercek: Ahh of course, thanks for that pickup!

Sometimes I get carried away with the double-dot notation.

joeconwaystk:
:+1: Thank you for contributing!

PR #629, opened 2019-01-31
Get client_id and client_secret from form data

edlman:
If the client_id and client_secret is not send in auth header, try to get it from form data. Some tools and clients send it this way. Eg. SoapUI.

Signed-off-by: Martin Edlman <ac@an.y-co.de>


edlman:
Hello,

I fully understand your concerns and I don't require the pull request to be
accepted. As I use SoapUI I need to solve the issue. I asked Smartbear
(SoapUI developers) to fix this, but until then I made this patch. I'll
possibly give a try to some other tool you mentioned.
Thank you for the reply and for the great work on Aqueduct!

Kind regards, Martin

Dne čt 31. 1. 2019 19:11 uživatel Joe Conway <notifications@github.com>
napsal:

PR #631, opened 2019-02-19
Escape special characters for LIKE expression for query expressions. …

### LINKED ISSUE #630: QueryExpression.equalTo() doesn't match backslashes as literals
The handling of backslashes in string properties is inconsistent between insert and fetch operations. For inserts, backslashes are treated literally, but in fetch, backslashes are treated as escape sequences.

The following test case fails, which was very unexpected for me. (The test case uses the [table definition from your rest_postgres example](https://github.com/stablekernel/aqueduct_examples/blob/master/rest_postgres/lib/model/model.dart)).
```dart
test('can find object with backslash-escaped value', () async {
  final name = r'my\value';
  await Query.insertObject(app.context, Model()..name = name);
  final model = await (Query<Model>(app.context)
        ..where((m) => m.name).equalTo(name))
      .fetchOne();
  expect(model, isNotNull);
});
```

I assume the reason for this behaviour is that Postgres recognizes these escapes internally. Section 4.1.2.2. of [chapter 11](https://www.postgresql.org/docs/11/sql-syntax-lexical.html) of the Postgres documentation states:
> Any other character following a backslash is taken literally.

Which means that this test case fails as well:
```dart
test('backslashes are matched as literals', () async {
  await Query.insertObject(app.context, Model()..name = 'hello world');
  final model = await (Query<Model>(app.context)
        ..where((m) => m.name).equalTo(r'h\e\l\l\o\ \w\or\l\d'))
      .fetchOne();
  expect(model, isNull);
});
```

I found this behaviour confusing and would suggest to change it so that string values are always matched literally.
If the current behaviour is actually desired, I would like to see it documented explicitly, for example in the `.equalTo()` docs.

MarcusSmith:
…Add like query expression that does not escape special characters. Fixes #630

### DIFF
--- a/aqueduct/lib/src/db/postgresql/builders/expression.dart
+++ b/aqueduct/lib/src/db/postgresql/builders/expression.dart
@@ -16,7 +16,7 @@ class ColumnExpressionBuilder extends ColumnBuilder {
   String get defaultPrefix => "$prefix${table.sqlTableReference}_";
 
   QueryPredicate get predicate {
-    var expr = expression;
+    final expr = expression;
     if (expr is ComparisonExpression) {
       return comparisonPredicate(expr.operator, expr.value);
     } else if (expr is RangeExpression) {
@@ -28,7 +28,8 @@ class ColumnExpressionBuilder extends ColumnBuilder {
     } else if (expr is StringExpression) {
       return stringPredicate(expr.operator, expr.value,
           caseSensitive: expr.caseSensitive,
-          invertOperator: expr.invertOperator);
+          invertOperator: expr.invertOperator,
+          allowSpecialCharacters: expr.allowSpecialCharacters);
     }
 
     throw UnsupportedError(
@@ -37,8 +38,8 @@ class ColumnExpressionBuilder extends ColumnBuilder {
 
   QueryPredicate comparisonPredicate(
       PredicateOperator operator, dynamic value) {
-    var name = sqlColumnName(withTableNamespace: true);
-    var variableName = sqlColumnName(withPrefix: defaultPrefix);
+    final name = sqlColumnName(withTableNamespace: true);
+    final variableName = sqlColumnName(withPrefix: defaultPrefix);
 
     return QueryPredicate(
         "$name ${ColumnBuilder.symbolTable[operator]} @$variableName$sqlTypeSuffix",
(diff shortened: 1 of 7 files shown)


aqueduct/lib/src/db/postgresql/builders/expression.dart
```diff
@@ -114,4 +117,8 @@ class ColumnExpressionBuilder extends ColumnBuilder {
     return QueryPredicate("$n $operation @$variableName$sqlTypeSuffix",
         {variableName: matchValue});
   }
+
+  String escapeLikeString(String input) {
+    return input.replaceAllMapped(RegExp("(\\\\|%|_)"), (Match m) => "\\${m[0]}");
```
pschiffmann: FYI: If a string literal doesn't contain escape sequences (`\n`, `\"`) or string interpolation, you can use [raw strings](https://www.dartlang.org/guides/language/language-tour#strings) if you don't want to escape backslashes: `r"(\\|%|_)"`

aqueduct/lib/src/db/query/matcher_expression.dart
```diff
@@ -104,14 +104,56 @@ class QueryExpression<T, InstanceType> {
...
     } else {
       expression = ComparisonExpression(value, PredicateOperator.notEqual);
     }
 
     return _createJunction();
   }
 
+  /// Adds a like expression to a query.
+  ///
+  /// A query will only return objects where the selected property is like [value].
```
pschiffmann: Maybe add a link to the Postgres documentation how a "like" relation is defined? Before I encountered this issue, I have never seen `%` or `_` used as wildcard characters.

aqueduct/test/managed_auth/managed_auth_storage_test.dart
```diff
@@ -261,6 +261,15 @@ void main() {
       } on AuthServerException {}
     });
 
+    test("Create token fails with wildcard password", () async {
+      try {
+        await auth.authenticate(createdUser.username, "%",
+            "com.stablekernel.app1", "kilimanjaro");
+        expect(true, false);
+        // ignore: empty_catches
+      } on AuthServerException {}
```
pschiffmann: This is equivalent to:
```dart
expect(
    () => auth.authenticate(createdUser.username, "%", "com.stablekernel.app1", "kilimanjaro"),
    throwsA(const TypeMatcher<AuthServerException>()));
```

MarcusSmith: I'm going to leave this one as it is to match the other tests in this file, but more importantly to keep the method under test separate from the expectations (keeps the tests easier to understand and debug)

pschiffmann: Yes, you're probably right ... my version isn't any simpler, only better code-golfed :)

MarcusSmith: I do love code golf :)

aqueduct/test/db/postgresql/matcher_test.dart
```diff
@@ -81,6 +107,107 @@ void main() {
...
+      results = await q.fetch();
+      expect(results.length, 0);
+    });
+
+    test("case insensitive default", () async {
+      var q = Query<TestModel>(context)
+        ..where((o) => o.email).like("0@A%", caseSensitive: false);
+      var results = await q.fetch();
+      expect(results.length, 1);
+      expect(results.first.id, 1);
```
joeconwaystk: Expect email here instead of id, no guarantee on id

PR #632, opened 2019-02-20
Add fromDynamic static function on ManagedSet

MarcusSmith:


aqueduct/lib/src/db/managed/set.dart
```diff
@@ -28,6 +28,11 @@ class ManagedSet<InstanceType extends ManagedObject> extends Object
 
   /// Creates a [ManagedSet] from an [Iterable] of [InstanceType]s.
   ManagedSet.from(Iterable<InstanceType> items) {
+    _innerValues = items.toList();
+  }
+
+  /// Creates a [ManagedSet] from an [Iterable] of [dynamic]s.
+  ManagedSet.fromDynamic(Iterable<dynamic> items) {
     _innerValues = items.toList().cast();
```
joeconwaystk: don't use cast because it is lazy. the colelction could still contain non-InstanceType elements, you won't find out until they are accessed. Instead, iterate over each item and add it to _innerValues

aqueduct/test/db/model_test.dart
```diff
@@ -64,14 +64,14 @@ void main() {
       reflect(user).setField(#name, 1);
 
       expect(true, false);
-    } on ValidationException catch (e) {
-      expectError(e, contains("invalid input value for 'name'"));
+    } on TypeError catch (e) {
```
joeconwaystk: let me sync with you on this

PR #639, opened 2019-02-28
Add a 204 No Content convenience constructor to Response

jadengis:
I was working with these `Response` objects and noticed there was no named constructor for a 204 No Content response, so I added one.

I know a lot of API design patterns use 204 for deletes, and figured adding this as a convenience constructor will we helpful for people who want to use 204, and want to maintain a consistent feel to their code.

I checked the `response_test.dart` file to see if I should add any tests for this, but it seems like a lot of the other constructors aren't tested, so I decided against adding one.


joeconwaystk:
👍

PR #648, opened 2019-03-23
Allows different authentication processes by scope

MoacirSchmidt:
This little change will allow different authentication processes per scope.

You can override authentication process and create another one based on scope extending class AuthServerDelegate and overriding getResourceOwner method.


joeconwaystk:
I'm not sure what this PR intends to do, and it should include tests to demonstrate and validate that the changes have the expected behavior. That being said, at a quick glance, I don't think this solution is appropriate and the behavior is already available by following these instructions: https://aqueduct.io/docs/auth/auth_scopes/#defining-permissible-scope

PR #660, opened 2019-04-19
Error in section Handling HTTP Requests in Path Variable Bindings

adoniaspp:
The code that is in the 'Path Variable Bindings' section:

```
class CityController extends ResourceController {
  
@Operation.get('id')
 Future<Response> getCityByID(@Bind.query('id') String cityID) async {
        return new Response.ok(cities.where((c) => c.id == cityID).toList());
 }
}
```

However I believe the correct one would be:

```
class CityController extends ResourceController {
  
@Operation.get('id')
  Future<Response> getCityByID(@Bind.path('id') String cityID) async {
        return new Response.ok(cities.where((c) => c.id == cityID).toList());
  }
}
```


joeconwaystk:
Is this a documentation change? If so, the PR should be from your docs/source branch to this repo's docs/source branch. The master branch is code only.

adoniaspp:
Sorry, I'll do it the right way.

PR #662, opened 2019-04-25
Allow `aqueduct db generate` command to support naming migration file (#554)

### LINKED ISSUE #554: Allow `aqueduct db generate` command to support naming migration file
One should be able to run something like `aqueduct db generate <migration name>` or  `aqueduct db generate --name <migration name>` which would generate `<incremented num>_<migration name>.migration.dart`.

Related file: `https://github.com/stablekernel/aqueduct/blob/f2eac278579556ac57dbac3e620248044fcce71c/aqueduct/lib/src/cli/commands/db_generate.dart`

Thoughts, @joeconwaystk ?

deathman92:
Fixes #554


joeconwaystk:
Great PR - lgtm.

PR #668, opened 2019-05-05
formatting and port number fix

suragch:
Aqueduct is running on port 8888, not 8080.


joeconwaystk:
thanks for the edits!

PR #684, opened 2019-05-15
Update validations.md

suragch:
Fixed minor grammar errors.

As a side note (I didn't make any changes regarding it and I'm not going to bother opening an issue), isn't a partial PUT non-RESTful? Shouldn't PATCH be used instead?

https://stackoverflow.com/questions/19732423/why-isnt-http-put-allowed-to-do-partial-updates-in-a-rest-api

### DIFF
--- a/source/docs/db/validations.md
+++ b/source/docs/db/validations.md
@@ -17,7 +17,7 @@ class _Tweet {
 
 ### Built-in Validators
 
-There are a handful of built-in validations for common operations. For example, it is common to apply a regular expression to a value to ensure it formatted correctly or restrict the possible values to a list of available options. Common validators are available as named constructors f the `Validate` class. Here is an example:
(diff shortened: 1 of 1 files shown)


joeconwaystk:
In response to the thoughts on PUT: yes, the HTTP spec says a PUT should replace an entire resource (or create one). You can also PUT on a subresource  - i.e. `PUT /users/1/name` - to do partial updates. PATCH has some onerous requirements that aren't quite as straightforward as 'just send a partial object' (JSON Merge Patch is pretty close though: https://tools.ietf.org/html/rfc7386).

The problem that a full representation PUT solves is a complex one. The comment about PUT might be poor info for a new developer and may be worth removing.

suragch:
Ah, ok. I don't think you need to necessarily change anything about the
wording regarding PUT. Once I read the docs about how it works it made
sense.

PR #690, opened 2019-05-21
Add column name property

axellebot:
Just a simple add to customize column name :)

Example :
```dart
class User{
  @Column(name: 'this_is_my_column')
  String username;
}
```

### DIFF
--- a/aqueduct/lib/src/compilers/orm/property_builder.dart
+++ b/aqueduct/lib/src/compilers/orm/property_builder.dart
@@ -2,8 +2,8 @@ import 'dart:mirrors';
 
 import 'package:aqueduct/src/compilers/orm/data_model_builder.dart';
 import 'package:aqueduct/src/compilers/orm/entity_builder.dart';
-import 'package:aqueduct/src/compilers/orm/validator_builder.dart';
 import 'package:aqueduct/src/compilers/orm/entity_mirrors.dart';
+import 'package:aqueduct/src/compilers/orm/validator_builder.dart';
 import 'package:aqueduct/src/db/managed/managed.dart';
 import 'package:aqueduct/src/db/managed/relationship_type.dart';
 import 'package:aqueduct/src/utilities/mirror_helpers.dart';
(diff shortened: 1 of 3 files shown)


joeconwaystk:
Thanks @axellebot. Unfortunately, this feature is more complex and requires a lot more testing than provided here. Tests (and code) need to address database migrations, return and insertion value mapping for database queries, and ensure that any serializable interfaces remain unchanged regardless of the underlying column name. The column name also needs to be validated to ensure it is a valid column name (e.g. no special characters).

joeconwaystk:
Yes, much of the reason alternative column naming has not yet been implemented is due to the features that surround it. A solution needs to account for all of these related, unimplemented scenarios.

joeconwaystk:
And any missing tests you see for other parts of the framework would be great to add as well!

joeconwaystk:
missing tests should be added. I don’t understand why you would want to add more untested features and create more issues when there is a clear path to resolving both issues.

joeconwaystk:
I am positive that this feature is more complex and requires additional effort and thought - closing this PR for now. FWIW, finding a mistake (the missing tests for table names) and using it as a justification for making another mistake (not testing column naming) that is known beforehand isn’t a net positive for the project. 

 A migration will fail with an appropriate error message if someone tries to use rename column. This method will not be automatically generated by a migration.

joeconwaystk:
Yes, PropertyBuilder is the “compiler” for properties. ManagedPropertyDescription is the runtime info for that property. You found the right spot for naming columns, but each application of that information needs to be accounted for. For example, the intermediate data model format (Schema types, eg SchemaColumn) does not have test requirements for when a property and column can be named differently. To ensure that, as the implementation changes, that this will continue to work it’d be great to categorize where all similar assumptions are and make them requirements by adding test. This is especially important now that we are going to provide runtimes for JIT and AOT. 

I’d use the IDE to follow code paths related to column names to find all usages first. This is definitely s desired feature, but having written a lot of the ORM layer, I am positive I’ve made more mistakes that might come to light with this change. The safest path forward is just to make sure we’ve accounted for this and captured assumptions as tests.

PR #723, opened 2019-08-13
Fix ManagedAuthDelegate.getAllowedScopes() check in AuthServer

pschiffmann:
Hey!

I found a bug where I couldn't create an OAuth token with less scope than that of the client I created it with. For example, if I have a client with scope `user`, I couldn't create a token with scope `user.self`.

The bug was caused by a switched up caller/callee: To check whether the client allowed scope is also allowed for this user, the call must be `clientAllowedScope.allowsScope(userScope)`, not `userScope.allowsScope(clientAllowedScope)`. IMHO, the name is really confusing and should be deprecated in favor of `isSubsetOrEqualTo`.

### DIFF
--- a/aqueduct/test/auth/authenticate_test.dart
+++ b/aqueduct/test/auth/authenticate_test.dart
@@ -210,6 +210,17 @@ void main() {
       expect(await auth.verify(token.accessToken) is Authorization, true);
     });
 
+    test("Can create with sub-scope of client scope", () async {
+      delegate.allowedScopes = [AuthScope("user")];
+      var token = await auth.authenticate(
+          createdUser.username,
+          InMemoryAuthStorage.defaultPassword,
+          "com.stablekernel.public.scoped",
+          null,
+          requestedScopes: [AuthScope("user.self")]);
+      expect(token.scopes, [AuthScope("user.self")]);
+    });
+
     test("Cannot verify token that doesn't exist", () async {
       try {
         await auth.verify("nonsense");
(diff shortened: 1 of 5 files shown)


joeconwaystk:
Good find - I have not yet got a chance to dive into it too deeply, but I agree that the `allowsScope` is a bad name. In this PR, would you be able to rename `allowsScope` to `isSubsetOrEqualTo` project-wide, and then re-implement `allowsScope` to call `isSubsetOrEqualTo`, and deprecate `allowsScope`?

pschiffmann:
Done!

aqueduct/test/auth/authenticate_test.dart
```diff
@@ -210,6 +210,17 @@ void main() {
...
+      delegate.allowedScopes = [AuthScope("user")];
+      var token = await auth.authenticate(
+          createdUser.username,
+          InMemoryAuthStorage.defaultPassword,
+          "com.stablekernel.public.scoped",
+          null,
+          requestedScopes: [AuthScope("user.self")]);
+      expect(token.scopes, [AuthScope("user.self")]);
+    });
+
```
joeconwaystk: Can you add one more test case that checks the negative case? Such that if the delegate allows scope `user:things`, and you try and request scope `user`, that it fails to grant a token (the requested scope is greater than the allowed scope)

pschiffmann: Good objection, done.

PR #730, opened 2019-09-01
Remove deprecated --no-packages-dir flag from cli create

### LINKED ISSUE #729: On dart 2.4.1 the "--packages-dir" flag for "pub get" is no longer used and does nothing
On dart 2.4.1, during project creation, a deprecation message is raised about the "--packages-dir" being deprecated.

```sh
$ pub get --no-packages-dir
The --packages-dir flag is no longer used and does nothing.
Resolving dependencies...
Got dependencies!
```

Are there versions of dart that should be supported that this flag is useful for? Or can it be assumed the newest dart will be used? I was wondering if I could update this part of project creation or if I need to handle versions of dart pre 2.4.

Thanks!

mhelmetag:
Fixes #729 

Just removed a flag that was deprecated back in August 2018 that's used in the cli for creating new projects.

### DIFF
--- a/aqueduct/lib/src/cli/commands/create.dart
+++ b/aqueduct/lib/src/cli/commands/create.dart
@@ -69,7 +69,7 @@ class CLITemplateCreator extends CLICommand with CLIAqueductGlobal {
     }
 
     displayInfo(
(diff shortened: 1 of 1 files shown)


joeconwaystk:
looks good to me. thanks. will merge once Appveyor flaky test passes

PR #758, opened 2020-01-01
Simple doc rename class change

rlweb:


joeconwaystk:
Can you send a pull request for docs/source instead of master?

PR #760, opened 2020-01-15
Add accept filter for serializable read

Reductions:
Added accept filter option to read method of serializable interface. The way it is done is backward compatible (if no accept filters are given the behavior stays the same as before).


joeconwaystk:
I agree with this being useful. I'll be able to review this in more detail later this week. The failing tests are due to a change in `package:http`, so nothing for you to do there yet. I will get a fix into master for those tests and you'll be able to rebase/merge this PR.

Reductions:
Thanks for the feed back.

joeconwaystk:
Got tests back to passing and merged a big PR. It created some conflicts in this PR, but they should be easy to resolve. LMK if you need help.

PR #773, opened 2020-01-28
Scope authorization in ManagedObjectController 

Reductions:
- [x] Allow scope authorization to be used for the different actions.
  - [x] Implement the feature
  - [x] Add test


aqueduct/lib/src/http/managed_object_controller.dart
```diff
@@ -344,6 +376,21 @@ class ManagedObjectController<InstanceType extends ManagedObject>
     return didFindObjects(results);
   }
 
+  void _authorizeForScopes(List<String> scopes) {
+    if (scopes != null) {
```
joeconwaystk: Can you change this to ```if (scopes == null) { 
  return; 
}```

aqueduct/lib/src/http/managed_object_controller.dart
```diff
@@ -344,6 +376,21 @@ class ManagedObjectController<InstanceType extends ManagedObject>
...
+      if (request.authorization == null) {
+        throw Response.serverError();
```
joeconwaystk: It'd be good to log (severe or error?) this situation since it'd be hard to debug otherwise.

aqueduct/lib/src/http/managed_object_controller.dart
```diff
@@ -344,6 +376,21 @@ class ManagedObjectController<InstanceType extends ManagedObject>
...
+      }
+
+      if (!AuthScope.verify(
```
joeconwaystk: Can you extract the result of this boolean expression to a variable to make it more readable?

aqueduct/test/http/subclass_managed_object_controller_with_scopes_test.dart
```diff
@@ -0,0 +1,199 @@
...
+
+  String name;
+  DateTime createdAt;
+}
+
+class SubclassWithScopes extends ManagedObjectController<TestModel> {
+  SubclassWithScopes(ManagedContext context) : super(context);
+
+  @override
+  List<String> get findObjectWithIdScopes => ['get'];
```
joeconwaystk: Can you provide a test scenario for the empty and multiple cases of acceptable scope?

aqueduct/lib/src/http/managed_object_controller.dart
```diff
@@ -108,6 +115,11 @@ class ManagedObjectController<InstanceType extends ManagedObject>
     }
   }
 
+  /// Used to authorize POST /<name> action
+  ///
+  /// You may override it to return the required scopes for the the actions to be authorized
+  List<String> get createNewObjectScopes => null;
```
joeconwaystk: I think there is a better interface to this feature than these 5 separate overridable fields for two reasons. 

1. You can use the ManagedObjectController without subclassing (and this is the intended use case).
2. The five separate fields seem dispersed... collapsing them into a single structure that has similar properties might work (e.g. `controller.scopes.create = [...]`)

itsjoeconway (changes requested):

joeconwaystk:
Good feature add, some small feedback items and it should be good to go. Love the tests!

aqueduct/lib/src/http/managed_object_controller.dart
Reductions: Good suggestion!
Would you be OK if I put then as optional field(fields) in the  constructor?

joeconwaystk: I think that makes sense, as long as it doesn't create a breaking change where existing managed object controller subclasses require a non-default constructor.

Reductions:
I may split the second part in a different pull request (The one about being able to filter the body decode for `POST` and `PUT`).

PR #799, opened 2020-03-30
Joins are taken in account for reduce queries

Reductions:


joeconwaystk:
Thanks! Would you be able to add a test to `test/db/postgresql/aggregate_function_test.dart` that runs the success case of this change?

The file `package:aqueduct/src/dev/model_graph.dart` has test ManagedObjects that have relationships. (This particular test's `setUp` method must be altered to include those models once the file is imported).

PR #855, opened 2020-06-03
support dart2js to run my web service in any host

mahmoudsalah37:
aqueduct support only some host 
i need to publish my web serice any where.


jairoFernandez:
maybe if you split the commits, is easier to review the PR

PR #877, opened 2020-07-26
Async support for custom validators

joaovsq:
This is useful when it is necessary to perform I/O operations to validate the data (a dns lookup, for example).


aqueduct/lib/src/db/managed/validation/managed.dart
```diff
@@ -15,20 +15,20 @@ class ManagedValidator {
...
   /// This method does not invoke [ManagedObject.validate] - any customization provided
   /// by a [ManagedObject] subclass that overrides this method will not be invoked.
-  static ValidationContext run(ManagedObject object,
-      {Validating event = Validating.insert}) {
+  static Future<ValidationContext> run(ManagedObject object,
+      {Validating event = Validating.insert}) async {
     final context = ValidationContext();
 
-    object.entity.validators.forEach((validator) {
+    for (var validator in object.entity.validators) {
```
Reductions, line 22: I think here it would be better to do:
```dart
final validations = object.entity.validators.map(...
);
await Future.wait(validations);
```

joaovsq: I forgot to consider backwards compatibility, made a adjustment.

joaovsq:
> There is one general problem with this approach. The change is not backward compatible.
> 
> Maybe it would be better to create a different method (for example asyncValidate()).

I forgot to consider backwards compatibility, made a adjustment.

joeconwaystk:
Thank you for taking the time to write this. 

I don't think this is the right solution - it forces a consumer of validation behavior to account for asynchronous operations. The overhead and constraints this imposes on validation consumers is not worth it to the majority of use cases. The purpose of a validation is to run on the data stored within a managed object. Validating a DNS lookup, for example, is not something that needs to happen within the model layer - an application's logic can do that prior to creating a managed object.

Issue #110, opened 2016-11-03
other db's ?


joeblew99:
really nice framework.

anyone know if there is support for sql dialects, so that other databases can be targeted ?

maybe this is envisaged on the roadmap ?

joeconwaystk:
To add support for another database, PersistentStore must be implemented. MySQL would likely be relatively straightforward, as it would be really similar to the Postgres implementation. There are a few places where things got tangled - specifically with DB migration and tests - where we would have to make some changes. The long term goal is to support other DBs as packages outside of Aqueduct.

joeblew99:
@joeconwaystk 
Sounds like the proper way to do it might be to introduce an actual dialects class, so that the PeristentStore and the Migration can both use them ?

joeconwaystk:
Not having a lot of experience with Hibernate, it looks like dialects are doing translations a lot closer to the syntax, whereas the goal for persistent store is to do those translations by behavior. Not saying that is going to work long-term, but that is the goal.

joeblew99:
i used NHibernate for years.
https://github.com/nhibernate/nhibernate-core/tree/master/src/NHibernate/Dialect

If you even want a source of already worked out dialects, that i think might be the best source.
Just saying..
Shall we leave this open ?

joeconwaystk:
You can use any of the Dart MySQL drivers with Aqueduct, you just can't use the ORM features like Query and ManagedObject. Same goes for MongoDB or any other database.

MoacirSchmidt:
Unfortunately, you cannot use oAuth classes also since the strongly depend on database (postgres) classes.

joeconwaystk:
This isn't accurate. The types `AuthServer`, `AuthController`, `AuthRedirectController`, `AuthToken`, `AuthClient`, `Authorizer` and `Authorization` have no dependency or even relationship with PostgreSQL or the ORM.

The abstract interface `AuthDelegate` is overridden for storage/retrieval for auth-related tasks. There is one provided implementation that uses the ORM (and therefore PostgreSQL). You can provide your own implementation for that one type, and still use all of the classes above without any changes.

MoacirSchmidt:
I would really like to extend the AuthDelegate interface (or another ones) to use MySQL. Could you please guide me to some documentation or source code to achieve that?

joeconwaystk:
Override this class in the docs, and pass instance of it to your AuthServer constructor: 

[...]

PR #880, opened 2020-08-05
 Added support for MySQL database

maurobotta:
Hi

Aqueduct support only one database, to add mysql it is very  great feature

Read this link, too
https://github.com/oldshen/aqueduct_mysql


sdxv:
Joe Conway has has previously stated that the support for additional databases should be provided via 3rd party packages.
https://github.com/stablekernel/aqueduct/issues/110#issuecomment-258059846
Granted the comment is old, but I agree with it. Adding additional db support into the aqueduct package adds extra bloat and maintenance when it can be offered cleanly in it's own package.

Issue #811, opened 2020-04-13
SSL support for aqueduct db upgrade


timosettels:
This is possibly more of a documentation issue rather than a feature request in case it is already implemented:

As many hosting providers require SSL to connect to production databases, the migration tool should support SSL connections. Currently 'aqueduct db upgrade --connect ...' does not seem to be configurable to establish a SSL connection. Is this correct? Could this be implemented?

Best regards and awesome work so far!

Reductions:
You mean when you used `aqueduct db upgrade --use-ssl --connection ...` it doesn't work?

timosettels:
Yes exactly. When i enter 
`aqueduct db upgrade --use-ssl --connect postgres://...`
The command takes a little longer to execute than without the `--use-ssl`-Flag but ultimately also fails: `*** There was an error connecting to the database...`.
The exact same connection data works with the aqueduct `PostgreSQLPersistentStore` though, but only when i specify 

`PostgreSQLPersistentStore.fromConnectionInfo(
      ...
      useSSL: true,
    );`

Thats why i assumed it is a SSL issue.

Aldo111:
Thank you for opening this issue! I've been struggling with getting SSL mode to work for db migration on a hosted db instance. I didn't even know that --use-ssl was an option for the upgrade as I spent ages looking for it.

Using --use-ssl now results in the same error as above "There was an error.."

Aldo111:
I tried this with a different hosted postgres instance (was previously using AWS). With digitalocean, I was able to make it a little deeper before it failed again:

`-- Aqueduct CLI Version: 3.2.1
-- Aqueduct project version: 3.2.1
SELECT versionNumber, dateOfUpgrade FROM _aqueduct_version_pgsql ORDER BY dateOfUpgrade ASC
HERE
-- Updating to version 1 on new database...
    PostgreSQL connecting, <my db>.
*** There was an error connecting to the database '<my db>'. Reason: unable to connect to database.
`

Aldo111:
I pulled a stacktrace: https://pastebin.com/8puLt9Uu

joeconwaystk:
Thanks for everyone's input. There is a bug in this command where the `--use-ssl` flag does not actually toggle whether the connection is SSL: https://github.com/stablekernel/aqueduct/blob/16fc395f24d5ebde0c35b8d76caa8add3c25ef43/aqueduct/lib/src/cli/scripts/run_upgrade.dart#L43.

[...]

PR #894, opened 2020-09-13
Fix non ssl connection in aqueduct db upgrade even if --use-ssl is used.

NayanJD:
The cli command aqueduct db upgrade was not able to connect to SSL only postgres instances while running migration scripts. Issue #811 

This PR changes src/cli/scripts/run_upgrade.dart -> RunUpgradeExecutable.execute to use SSL if --use-ssl is used while aqueduct db upgrade.


lee-40square:
@joeconwaystk what do you think about this PR? Something that could be merged soon? :-)

PR #901, opened 2020-10-01
Fixing the way aqueduct TestRequest serializes list query parameters

Reductions:
Up until now list query parameters were serialized in such a way:
```
url?query=[1, 2]
```
This is not the way aqueduct expects list parameters to be sent with that pull request the way lists are serialized is:
```
url?query=1&query=2
```

### DIFF
--- a/aqueduct_test/lib/src/request.dart
+++ b/aqueduct_test/lib/src/request.dart
@@ -76,12 +76,19 @@ class TestRequest {
 
     var url = _baseUrl.resolve(actualPath).toString();
     if ((query?.length ?? 0) > 0) {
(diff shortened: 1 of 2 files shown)


ali-thowfeek:
@Reductions  I was trying to write a test like
```dart
harness.publicAgent.get("/search", query: {"query": "1", "query": "2"});
```
But soon realized that it's impossible do that in dart with the map containing duplicate keys.
would this PR fix it? where i would be able to write
```dart
harness.publicAgent.get("/search", query: {"query": ["1", "2"]});
```

Reductions:
Yes it would. But it would not be merged as the framework is no longer supported. Run away while you have an option.

PR #909, opened 2020-11-05
use int.tryParse instead in heroes example

tusharsadhwani:
since the endpoint is `/heroes/[:id]`, endpoints like `/heroes/abc` are technically valid, but that currently throws an HTTP 500 error (handler throws an exception), where it should really be a 404 error.

Using `int.tryParse` instead of `int.parse` is an easy fix for this issue.


tusharsadhwani:
Nevermind, I should read the docs before making PRs :)

PR #955, opened 2021-03-08
Nullsafety

j4qfrost:
If anyone would like to help...


aqueduct/lib/managed_auth.dart
```diff
@@ -108,47 +108,47 @@ class _ManagedAuthToken {
...
-  DateTime expirationDate;
+  DateTime? expirationDate;
 
   /// The resource owner of this token.
   ///
   /// [ResourceOwnerTableDefinition] must be implemented by some [ManagedObject] subclass in an application.
   /// That subclass will be the 'owner' of tokens. See [ResourceOwnerTableDefinition] for more details.
   @Relate.deferred(DeleteRule.cascade, isRequired: true)
-  ResourceOwnerTableDefinition resourceOwner;
+  late ResourceOwnerTableDefinition resourceOwner;
```
Reductions, line 143: `late` is nonsence when used in a `Model`.

Reductions (commented):
3 things I want to say:
1. Your pull request is never going to be merged.
2. You don't understand how to framwork works under the hood, and the changes you are making are mostly mechanical.
3. When you did find and replace of `new` you've deleted a lot of the documentation.

Advice if you want to support the framework fork it rename it along with all packages that are in the stablekernel repo. Then republish.

aqueduct/lib/managed_auth.dart
j4qfrost: It's better than nothing, I realize this is a very naive migration, and that's why I'm asking for help. I made this a draft, not a real pull request. If you're getting annoyed by notifications, you can turn them off. If this doesn't get merged then it doesn't get merged.

